PR #47, opened 2020-11-20
Add docs generation script.

Lukasa:
Motivation:

Documentation is a real help when using a project, we should be
generating it.

Modifications:

- Ported docs script from SwiftNIO.

Result:

We can provide API documentation.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,7 +26,7 @@ RUN mkdir -p $HOME/.tools
 RUN echo 'export PATH="$HOME/.tools:$PATH"' >> $HOME/.profile
 
 # script to allow mapping framepointers on linux (until part of the toolchain)
-RUN wget -q https://raw.githubusercontent.com/apple/swift/master/utils/symbolicate-linux-fatal -O $HOME/.tools/symbolicate-linux-fatal
+RUN wget -q https://raw.githubusercontent.com/apple/swift/main/utils/symbolicate-linux-fatal -O $HOME/.tools/symbolicate-linux-fatal
 RUN chmod 755 $HOME/.tools/symbolicate-linux-fatal
 
 # swiftformat (until part of the toolchain)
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN mkdir -p $HOME/.tools
 RUN echo 'export PATH="$HOME/.tools:$PATH"' >> $HOME/.profile
 
 # script to allow mapping framepointers on linux (until part of the toolchain)
-RUN wget -q https://raw.githubusercontent.com/apple/swift/master/utils/symbolicate-linux-fatal -O $HOME/.tools/symbolicate-linux-fatal
+RUN wget -q https://raw.githubusercontent.com/apple/swift/main/utils/symbolicate-linux-fatal -O $HOME/.tools/symbolicate-linux-fatal
 RUN chmod 755 $HOME/.tools/symbolicate-linux-fatal
```
tomerd, line 30: we can probably just get rid of this all-together, we do not need it

scripts/generate_docs.sh
```diff
@@ -0,0 +1,127 @@
...
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -ex
+
+my_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+root_path="$my_path/.."
+version=$(git describe --abbrev=0 --tags || echo "0.0.0")
+modules=( NIOSSH )
```
tomerd, line 21: I see there a re a f ew other modules in the repo, is including only NIOSSH in the docs intentional?

tomerd (approved):
looks good, but a couple of questions

scripts/generate_docs.sh
Lukasa: Yes, the other modules are binaries.

docker/Dockerfile
Lukasa: Done.

PR #52, opened 2020-11-20
Resolve stalls in the example client.

Lukasa:
Motivation:

When running the example client on extremely low latency connections it
can occasionally stall, printing nothing to stdout before exiting. This
is the result of a timing window where we receive the data before the
pipe bootstrap has been configured and our glue handlers are not in
place.

There is also another timing window where we block reads on the SSH
Child Channel because the glue handler asked if the partner was writable
before the partner was inserted into the pipe channel.

Modifications:

- Cleanup both timing windows.

Result:

The SSH client example will be substantially more reliable.

### DIFF
--- a/Sources/NIOSSHServer/GlueHandler.swift
+++ b/Sources/NIOSSHServer/GlueHandler.swift
@@ -72,6 +72,12 @@ extension GlueHandler: ChannelDuplexHandler {
 
     func handlerAdded(context: ChannelHandlerContext) {
         self.context = context
+
+        // It's possible our partner asked if we were writable, before, and we couldn't answer.
+        // Consider updating it.
+        if context.channel.isWritable {
+            self.partner?.partnerBecameWritable()
+        }
     }
 
     func handlerRemoved(context: ChannelHandlerContext) {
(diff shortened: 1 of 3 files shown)


Sources/NIOSSHServer/GlueHandler.swift
```diff
@@ -72,6 +72,12 @@ extension GlueHandler: ChannelDuplexHandler {
 
     func handlerAdded(context: ChannelHandlerContext) {
         self.context = context
+
+        // It's possible our partner asked if we were writable, before, and we couldn't answer.
+        // Consider updating it.
+        if context.channel.isWritable {
+            self.partner?.partnerBecameWritable()
```
Joannis, line 79: Is there a scenario where we became writable but the partner wasn't set yet? This also goes for `channelWritabilityChanged`.

Joannis: Seeing the current usages, and the fact that the class is internal, I think it's never going to happen. So it might be wise to make it implicitly unwrapped.

Lukasa: In principle the partner should always be set, but we do nil it out explicitly on `handlerRemoved` so it’s likely better safe than sorry here.

artemredkin (approved):

PR #53, opened 2020-11-20
Use `getpass` to retrieve password from user

helje5:
Don't show the user's password input in the terminal.
Except on Windows, everything goes there.


Lukasa:
@swift-server-bot add to whitelist

Lukasa (approved):
LGTM. We'll need to wait for a while to get the CI sorted out but I'm otherwise happy.

PR #58, opened 2020-11-27
Add a ChannelOption for a server to query the authenticated username.

jimstudt:
Some servers will need to know which user authenticated. This PR exposes that information through a `ChannelOption` named `SSHChildChannelOptions.Types. UsernameOption` which is a `String?` of the username if authenticated, otherwise nil.

The new code passes the `username` along with the chain of possible completions from an authentication request until it reaches the creation of the `.userAuthSuccess` message. This message is augmented with an associated `String?` to hold the username. Eventually, this message will arrive at the `SSHConnectionStateMachine`, where it will be recorded.

The `NIOSSHHandler` contains the `SSHConnectionStateMachine`, and there is a parent path reaching up from the `SSHChildHandler` to the `NIOSSHHandler`. This path is festooned with `var username:String?` properties to allow 
the `SSHChildHandler` to reach up and back down to get the username when the `UsernameOption` is requested.

I wonder a bit about adding the associated type to the `.userAuthSuccess` SSHMessage, but most of them have associated 
data already. I had to add explicit `nil` values wherever they are created without a username.

I could have stored the actual username directly in the `NIOSSHHandler`, but that would have required adding a per-message inspection in `NIOSSHHandler`. It was already being done in `SSHConnectionStateMachine`, but there isn't a parent link back up to set the value in `NIOSSHHandler`.

I didn't find a test which sets up a server and checks its behavior in order to write tests. I note that the other `ChannelOption`s are also untested and suspect that has something to do with it. My knowledge of NIO stops a little
short of understanding the `BackToBackEmbeddedChannel` test harness. I've been testing the PR from my SSHConsole 
project's `echo` server and it appears to work fine.

### DIFF
--- a/Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
+++ b/Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
@@ -53,4 +53,11 @@ extension SSHChildChannelOptions.Types {
 
         public init() {}
     }
+    
+    /// `UsernameOption` allows users to query the authenticated username of the channel.
+    public struct UsernameOption: ChannelOption {
+        public typealias Value = String?
+        
+        public init() {}
+    }
 }
--- a/Sources/NIOSSH/Child Channels/SSHChannelMultiplexer.swift
+++ b/Sources/NIOSSH/Child Channels/SSHChannelMultiplexer.swift
@@ -93,6 +93,9 @@ extension SSHChannelMultiplexer {
             self.erroredChannels.append(channelID)
         }
     }
+    
+    // The username which the server accepted in authorization
+    var username : String? { (delegate as? NIOSSHHandler)?.username }
 }
 
 // MARK: Calls from SSH handlers.
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -41,7 +41,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthRequest(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(Self.transform(_:)))
+            return .possibleFutureMessage(future.map{ Self.transform($0, username: message.username) } )
         } else {
             return .noMessage
         }
@@ -65,10 +65,10 @@ extension AcceptsUserAuthMessages {
         }
     }
 
-    private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
+    private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage, username: String?) -> SSHMultiMessage {
         switch result {
         case .success:
-            return SSHMultiMessage(.userAuthSuccess)
+            return SSHMultiMessage(.userAuthSuccess(username))
         case .failure(let message):
             return SSHMultiMessage(.userAuthFailure(message))
         case .publicKeyOK(let message):
--- a/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
+++ b/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
@@ -59,7 +59,10 @@ struct SSHConnectionStateMachine {
 
     /// The state of this state machine.
     private var state: State
-
+    
+    /// The username which authenticated to the server, if that has happened
+    private(set) var username: String? = nil
+    
     private static let defaultTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
         AES256GCMOpenSSHTransportProtection.self, AES128GCMOpenSSHTransportProtection.self,
     ]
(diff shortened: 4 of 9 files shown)


Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
```diff
@@ -53,4 +53,11 @@ extension SSHChildChannelOptions.Types {
 
         public init() {}
     }
+    
+    /// `UsernameOption` allows users to query the authenticated username of the channel.
+    public struct UsernameOption: ChannelOption {
+        public typealias Value = String?
+        
+        public init() {}
+    }
```
Lukasa, line 62: Can you add the associated `static let` above?

Sources/NIOSSH/Child Channels/SSHChannelMultiplexer.swift
```diff
@@ -93,6 +93,9 @@ extension SSHChannelMultiplexer {
             self.erroredChannels.append(channelID)
         }
     }
+    
+    // The username which the server accepted in authorization
+    var username : String? { (delegate as? NIOSSHHandler)?.username }
```
Lukasa, line 98: Rather than cast the delegate, can we add the username computed property to it?

Also, while we're here, a quick NIO style nit:

```suggestion
    var username: String? { (delegate as? NIOSSHHandler)?.username }
```

Sources/NIOSSH/SSHMessages.swift
```diff
@@ -35,7 +35,7 @@ enum SSHMessage: Equatable {
     case newKeys
     case userAuthRequest(UserAuthRequestMessage)
     case userAuthFailure(UserAuthFailureMessage)
-    case userAuthSuccess
+    case userAuthSuccess(String?)
```
Lukasa, line 38: We don't want to make this change. This enum stores strictly the messages, as they are represented on the wire. This message does not carry a username, so we shouldn't be attaching it here.

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
```diff
@@ -59,7 +59,10 @@ struct SSHConnectionStateMachine {
 
     /// The state of this state machine.
     private var state: State
-
+    
+    /// The username which authenticated to the server, if that has happened
+    private(set) var username: String? = nil
```
Lukasa, line 64: Rather than store this as a stored property here, we should make it a computed property that calls into a computed property on `State`. That computed property can then switch over the state and load it from the underlying state data structure.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
```diff
@@ -41,7 +41,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthRequest(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(Self.transform(_:)))
+            return .possibleFutureMessage(future.map{ Self.transform($0, username: message.username) } )
```
Lukasa, line 44: We probably need to amend the user auth state machine to keep track of the username for which the user has been authenticating. We can do this in a circular buffer (to maintain ordering). When we complete the user auth we know what the user authenticated for and can store this on the relevant state.

As with the connection state machine, we should also expose a computed property that lets the connection state machine ask what the username was. It can then pass that through to other states.

Lukasa (commented):
Thanks for this patch! This looks like a good start! I've left a few notes in the diff about how we might want to take this forward.

Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
jimstudt: Done

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
jimstudt: I'm trying a few things here, but I don't think ordering will work. The many outstanding public key requests to the authentication delegate can complete in any order. I've got cleaner implementation going, but fundamentally it still involves pairing the ultimate future with a username.

Sources/NIOSSH/SSHMessages.swift
jimstudt: Good enough, it felt dirty when I did it. I have another strategy for connecting from the escaped closure world of the authentication delegate to the state structures.

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
jimstudt: This almost worked, but ran into trouble because an escaping closure cannot keep a mutating self reference. There just isn't any information to tie the successful conclusion of a authentication delegate to the username without tying it up in a closure or changing the delegate API which would be bad.

The next proposed solution introduces a group of mutable attributes on the SSHConnectionStateMachine which are in a class object, so can be handed around and updated as needed.

jimstudt (commented):
This touches a lot of files, but mostly that's just adding it to each state.

I noticed a couple comments about TODO work needing to limit the number of attempts for various activities, this 'connectionAttributes' area would be an easy place to implement a counter for those in the future.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
Lukasa: The user must complete them in order though, as the wire protocol has no way to associate each request with a response.

Joannis (commented):
Looks like a good set of changes to me!

Joannis:
@Lukasa can we merge this?

PR #62, opened 2020-12-21
Add support for custom public-private key authentication types

Joannis:
Notes:

I've implemented RSA in a separate module within this repo, to be split out of this repository before merging. It serves as an example implementation for this PR.

I believe there are a couple of API design choices, like protocol names, that should be improved. I'm happy to take any feedback, of course.

### DIFF
--- a/Sources/NIOSSH/Keys And Signatures/CustomKeys.swift
+++ b/Sources/NIOSSH/Keys And Signatures/CustomKeys.swift
@@ -1,0 +1,44 @@
+import Foundation
+import NIO
+
+public protocol NIOSSHSignatureProtocol {
+    static var signaturePrefix: String { get }
+    var rawRepresentation: Data { get }
+    
+    func write(to buffer: inout ByteBuffer) -> Int
+}
+
+internal extension NIOSSHSignatureProtocol {
+    var signaturePrefix: String {
+        Self.signaturePrefix
+    }
+}
+
+public protocol NIOSSHPublicKeyProtocol {
+    static var publicKeyPrefix: String { get }
+    var rawRepresentation: Data { get }
+    
+    func isValidSignature<D: DataProtocol>(_ signature: NIOSSHSignatureProtocol, for data: D) -> Bool
+    
+    func write(to buffer: inout ByteBuffer) -> Int
+}
+
+internal extension NIOSSHPublicKeyProtocol {
+    var publicKeyPrefix: String {
+        Self.publicKeyPrefix
+    }
+}
+
+public protocol NIOSSHPrivateKeyProtocol {
+    static var keyPrefix: String { get }
+//    var rawRepresentation: Data { get }
+    var publicKey: NIOSSHPublicKeyProtocol { get }
+    
+    func signature<D: DataProtocol>(for data: D) throws -> NIOSSHSignatureProtocol
+}
+
+internal extension NIOSSHPrivateKeyProtocol {
+    var keyPrefix: String {
+        Self.keyPrefix
+    }
+}
--- a/Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
+++ b/Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
@@ -88,12 +88,17 @@ extension NIOSSHPublicKey {
             return digest.withUnsafeBytes { digestPtr in
                 key.isValidSignature(sig, for: digestPtr)
             }
+        case (.rsa(let key), .rsa(let sig)):
+            return digest.withUnsafeBytes { digestPtr in
+                key.isValidSignature(sig, for: digestPtr)
+            }
         case (.certified(let key), _):
             return key.isValidSignature(signature, for: digest)
         case (.ed25519, _),
              (.ecdsaP256, _),
              (.ecdsaP384, _),
-             (.ecdsaP521, _):
+             (.ecdsaP521, _),
+             (.rsa, _):
             return false
         }
     }
@@ -110,12 +115,15 @@ extension NIOSSHPublicKey {
             return key.isValidSignature(sig, for: bytes.readableBytesView)
         case (.ecdsaP521(let key), .ecdsaP521(let sig)):
             return key.isValidSignature(sig, for: bytes.readableBytesView)
+        case (.rsa(let key), .rsa(let sig)):
+            return key.isValidSignature(sig, for: bytes.readableBytesView)
         case (.certified(let key), _):
             return key.isValidSignature(signature, for: bytes)
         case (.ed25519, _),
              (.ecdsaP256, _),
              (.ecdsaP384, _),
-             (.ecdsaP521, _):
+             (.ecdsaP521, _),
+             (.rsa, _):
             return false
         }
     }
@@ -132,12 +140,15 @@ extension NIOSSHPublicKey {
             return key.isValidSignature(sig, for: payload.bytes.readableBytesView)
         case (.ecdsaP521(let key), .ecdsaP521(let sig)):
             return key.isValidSignature(sig, for: payload.bytes.readableBytesView)
+        case (.rsa(let key), .rsa(let sig)):
+            return key.isValidSignature(sig, for: payload.bytes.readableBytesView)
         case (.certified(let key), _):
             return key.isValidSignature(signature, for: payload)
         case (.ed25519, _),
              (.ecdsaP256, _),
              (.ecdsaP384, _),
-             (.ecdsaP521, _):
+             (.ecdsaP521, _),
+             (.rsa, _):
             return false
         }
     }
@@ -150,6 +161,7 @@ extension NIOSSHPublicKey {
         case ecdsaP256(P256.Signing.PublicKey)
         case ecdsaP384(P384.Signing.PublicKey)
         case ecdsaP521(P521.Signing.PublicKey)
+        case rsa(Insecure.RSA.Signing.PublicKey)
         case certified(NIOSSHCertifiedPublicKey) // This case recursively contains `NIOSSHPublicKey`.
     }
 
@@ -164,6 +176,9 @@ extension NIOSSHPublicKey {
 
     /// The prefix of a P521 ECDSA public key.
     internal static let ecdsaP521PublicKeyPrefix = "ecdsa-sha2-nistp521".utf8
+    
+    /// The prefix of a RSA public key.
+    internal static let rsaPublicKeyPrefix = "ssh-rsa".utf8
 
     internal var keyPrefix: String.UTF8View {
         switch self.backingKey {
@@ -175,13 +190,15 @@ extension NIOSSHPublicKey {
             return Self.ecdsaP384PublicKeyPrefix
         case .ecdsaP521:
             return Self.ecdsaP521PublicKeyPrefix
+        case .rsa:
+            return Self.rsaPublicKeyPrefix
         case .certified(let base):
             return base.keyPrefix
         }
     }
 
     internal static var knownAlgorithms: [String.UTF8View] {
-        [Self.ed25519PublicKeyPrefix, Self.ecdsaP384PublicKeyPrefix, Self.ecdsaP256PublicKeyPrefix, Self.ecdsaP521PublicKeyPrefix]
+        [Self.ed25519PublicKeyPrefix, Self.ecdsaP384PublicKeyPrefix, Self.ecdsaP256PublicKeyPrefix, Self.ecdsaP521PublicKeyPrefix, Self.rsaPublicKeyPrefix]
     }
 }
 
@@ -197,12 +214,15 @@ extension NIOSSHPublicKey.BackingKey: Equatable {
             return lhs.rawRepresentation == rhs.rawRepresentation
         case (.ecdsaP521(let lhs), .ecdsaP521(let rhs)):
             return lhs.rawRepresentation == rhs.rawRepresentation
+        case (.rsa(let lhs), .rsa(let rhs)):
+            return lhs == rhs
         case (.certified(let lhs), .certified(let rhs)):
             return lhs == rhs
         case (.ed25519, _),
              (.ecdsaP256, _),
              (.ecdsaP384, _),
              (.ecdsaP521, _),
+             (.rsa, _),
              (.certified, _):
             return false
         }
@@ -224,8 +244,11 @@ extension NIOSSHPublicKey.BackingKey: Hashable {
         case .ecdsaP521(let pkey):
             hasher.combine(4)
             hasher.combine(pkey.rawRepresentation)
-        case .certified(let pkey):
+        case .rsa(let pkey):
             hasher.combine(5)
+            hasher.combine(pkey.rawRepresentation)
+        case .certified(let pkey):
+            hasher.combine(6)
             hasher.combine(pkey)
         }
     }
@@ -250,6 +273,8 @@ extension ByteBuffer {
         case .ecdsaP521(let key):
             writtenBytes += self.writeSSHString(NIOSSHPublicKey.ecdsaP521PublicKeyPrefix)
             writtenBytes += self.writeECDSAP521PublicKey(baseKey: key)
+        case .rsa(let key):
+            writtenBytes += self.writeRSAPublicKey(baseKey: key)
         case .certified(let key):
             return self.writeCertifiedKey(key)
         }
@@ -271,6 +296,8 @@ extension ByteBuffer {
             return self.writeECDSAP384PublicKey(baseKey: key)
         case .ecdsaP521(let key):
             return self.writeECDSAP521PublicKey(baseKey: key)
+        case .rsa(let key):
+            return self.writeRSAPublicKey(baseKey: key)
         case .certified:
             preconditionFailure("Certified keys are the only callers of this method, and cannot contain themselves")
         }
@@ -334,6 +361,15 @@ extension ByteBuffer {
         var writtenBytes = 0
         writtenBytes += self.writeSSHString("nistp521".utf8)
         writtenBytes += self.writeSSHString(baseKey.x963Representation)
+        return writtenBytes
+    }
+    
+    private mutating func writeRSAPublicKey(baseKey: Insecure.RSA.Signing.PublicKey) -> Int {
+        // For ssh-rsa, the format is public exponent `e` followed by modulus `n`
+        var writtenBytes = 0
+        writtenBytes += self.writeSSHString("ssh-rsa".utf8)
+        writtenBytes += self.writePositiveMPInt(baseKey.publicExponent.serialize())
+        writtenBytes += self.writePositiveMPInt(baseKey.modulus.serialize())
         return writtenBytes
     }
 
--- a/Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
+++ b/Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
@@ -88,7 +88,7 @@ extension NIOSSHPublicKey {
             return digest.withUnsafeBytes { digestPtr in
                 key.isValidSignature(sig, for: digestPtr)
             }
-        case (.rsa(let key), .rsa(let sig)):
+        case (.custom(let key), .custom(let sig)):
             return digest.withUnsafeBytes { digestPtr in
                 key.isValidSignature(sig, for: digestPtr)
             }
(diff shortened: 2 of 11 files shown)


Joannis:
@Lukasa please do review once you'e back from a well deserved vacation.

Lukasa:
This seems like a great start! The big missing link here seems to be user authentication, which I'm fine to omit for now as dealing simply with user and host keys will solve the 90% use-case handily. As this PR demonstrates, it's also _much_ easier to implement!

I'd be entirely happy with you productising this PR.

Joannis:
@Lukasa I've updated it all except the docs. Before I add the docs, what do you think of registering SSH public keys?

Lukasa:
I think that's going to be necessary. We can add a thread-safe registry allowing what amount to "plugins" for SSH key types.

Joannis:
I appearantly forgot to push my commit & ping you. Is this sufficient?

waylybaye:
Hello, any plan to merge the RSA keys support?

Lukasa:
@waylybaye This PR sort of got lost in the shuffle, but I'd happily review a version of this brought up-to-date with the current `main`.

waylybaye:
@Lukasa Great! The only thing blocking me to use nio-ssh is the lack of RSA support. Thank you and Joannis.

Lukasa:
This conflicts with the changes made in #98, can we clean that up?

Joannis:
@Lukasa is that the only requested/required change?

Lukasa:
Nope, it just cleans the diff up enough that I can do a proper review. :smile:

Joannis:
Done @Lukasa

Tests/NIOSSHTests/EndToEndTests.swift
```diff
@@ -435,6 +771,121 @@ class EndToEndTests: XCTestCase {
         XCTAssertEqual(self.channel.activeServerChannels.count, 1)
         #endif
     }
+    
+    func testCustomPublicKeyAlgorithms() throws {
+        CustomPublicKey.wasUsed = false
+        NIOSSHAlgoritms.register(publicKey: CustomPublicKey.self, signature: CustomSignature.self)
```
Lukasa, line 777: This kind of global state is really awkward for testing purposes because it makes it hard to remove. Can we consider registering things on a specific instance of configuration, rather than globally? That will help solve our various locking problems as well as our test issues.

Lukasa (commented):
Great, this is a really solid start! I've left some notes in the diff.

Sources/NIOSSH/Key Exchange/EllipticCurveKeyExchange.swift
```diff
@@ -16,33 +16,49 @@ import Crypto
...
-                                                initialExchangeBytes: inout ByteBuffer,
-                                                allocator: ByteBufferAllocator,
-                                                expectedKeySizes: ExpectedKeySizes) throws -> (KeyExchangeResult, SSHMessage.KeyExchangeECDHReplyMessage)
+    mutating func completeKeyExchangeServerSide(
+        clientKeyExchangeMessage message: ByteBuffer,
+        serverHostKey: NIOSSHPrivateKey,
+        initialExchangeBytes: inout ByteBuffer,
+        allocator: ByteBufferAllocator,
+        expectedKeySizes: ExpectedKeySizes
+    ) throws -> (KeyExchangeResult, NIOSSHKeyExchangeServerReply)
```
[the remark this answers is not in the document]
Joannis: This is definitely something worth discussing & exploring.

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -224,14 +246,48 @@ extension NIOSSHPublicKey.BackingKey: Hashable {
...
 
+extension NIOSSHPublicKey {
+    @discardableResult
+    public func write(to buffer: inout ByteBuffer) -> Int {
+        return buffer.writeSSHHostKey(self)
+    }
+    
+    @discardableResult
+    public func writeWithoutHeader(to buffer: inout ByteBuffer) -> Int {
+        return buffer.writeSSHHostKeyWithoutHeader(self)
```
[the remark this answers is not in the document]
Joannis: I don't quite need `writeWithoutHeader` in my use case, but the `write(to: )` function is required for implementing DH Group 14 SHA1.

Tests/NIOSSHTests/EndToEndTests.swift
Joannis: Do you mean in SSHClientConfiguration or SSHServerConfiguration?

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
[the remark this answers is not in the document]
Lukasa: It's necessary to write the host key in that context?

Tests/NIOSSHTests/EndToEndTests.swift
Lukasa: Sadly both, but we can use a single type stored in both places.

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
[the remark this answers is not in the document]
Joannis: Yes: https://datatracker.ietf.org/doc/html/rfc4253#section-8

Tests/NIOSSHTests/EndToEndTests.swift
Joannis: Yeah, cool! On the same page then

Sources/NIOSSH/Keys And Signatures/CustomKeys.swift
```diff
@@ -0,0 +1,109 @@
...
+internal extension NIOSSHSignatureProtocol {
+    var signaturePrefix: String {
+        Self.signaturePrefix
+    }
+}
+
+public enum NIOSSHAlgoritms {
+    public static func register(keyExchangeAlgorithm type: NIOSSHKeyExchangeAlgorithmProtocol.Type) {
+        if !customKeyExchangeAlgorithms.contains(where: { ObjectIdentifier($0) == ObjectIdentifier(type) }) {
+            customKeyExchangeAlgorithms.append(type)
```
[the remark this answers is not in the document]
Joannis: We can't, because we're not storing an instance. We're storing the metatype.

Sources/NIOSSH/Keys And Signatures/CustomKeys.swift
```diff
@@ -0,0 +1,109 @@
...
+    
+    public static func register(transportProtectionScheme type: NIOSSHTransportProtection.Type) {
+        if !SSHConnectionStateMachine.defaultTransportProtectionSchemes.contains(where: { ObjectIdentifier($0) == ObjectIdentifier(type) }) {
+            SSHConnectionStateMachine.defaultTransportProtectionSchemes.append(type)
+        }
+    }
+    
+    /// Registers a custom type tuple for use in Public Key Authentication.
+    public static func register<
+        PublicKey: NIOSSHPublicKeyProtocol,
```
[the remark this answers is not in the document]
Joannis: It's not necessary in the current implementation. The private keys are always used through their instances, and we won't ever need to acquire its meta type in order to engage a process. So unless we support some mechanism of loading a supported private key from the a blob, I don't foresee any need.

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -65,7 +65,7 @@ public struct NIOSSHPublicKey: Hashable {
 extension NIOSSHPublicKey {
     /// Verifies that a given `NIOSSHSignature` was created by the holder of the private key associated with this
     /// public key.
-    internal func isValidSignature<DigestBytes: Digest>(_ signature: NIOSSHSignature, for digest: DigestBytes) -> Bool {
+    public func isValidSignature<DigestBytes: Digest>(_ signature: NIOSSHSignature, for digest: DigestBytes) -> Bool {
```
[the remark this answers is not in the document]
Joannis: For verifying the exchange hash in custom key exchange algorithms.

Sources/NIOSSH/Keys And Signatures/CustomKeys.swift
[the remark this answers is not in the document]
Lukasa: Oh yeah, I see. That's fair enough then.

Sources/NIOSSH/Key Exchange/SSHKeyExchangeStateMachine.swift
```diff
@@ -71,16 +71,18 @@ struct SSHKeyExchangeStateMachine {
     private let role: SSHConnectionRole
     private var state: State
     private var initialExchangeBytes: ByteBuffer
-    private var protectionSchemes: [NIOSSHTransportProtection.Type]
+    private var transportProtectionSchemes: [NIOSSHTransportProtection.Type]
+    private var keyExchangeAlgorithms: [NIOSSHKeyExchangeAlgorithmProtocol.Type]
     private var previousSessionIdentifier: ByteBuffer?
 
-    init(allocator: ByteBufferAllocator, loop: EventLoop, role: SSHConnectionRole, remoteVersion: String, protectionSchemes: [NIOSSHTransportProtection.Type], previousSessionIdentifier: ByteBuffer?) {
+    init(allocator: ByteBufferAllocator, loop: EventLoop, role: SSHConnectionRole, remoteVersion: String, keyExchangeAlgorithms: [NIOSSHKeyExchangeAlgorithmProtocol.Type] = SSHKeyExchangeStateMachine.bundledKeyExchangeImplementations, transportProtectionSchemes: [NIOSSHTransportProtection.Type] = SSHConnectionStateMachine.bundledTransportProtectionSchemes, previousSessionIdentifier: ByteBuffer?) {
```
Lukasa, line 78: Is there any reason to default the keyExchangeAlgorithms and transportProtectionSchemes here?

Sources/NIOSSH/Key Exchange/SSHKeyExchangeStateMachine.swift
```diff
@@ -369,13 +377,13 @@ struct SSHKeyExchangeStateMachine {
...
         case .client:
-            clientAlgorithms = Self.supportedKeyExchangeAlgorithms
+            clientAlgorithms = role.keyExchangeAlgorithms.flatMap { $0.keyExchangeAlgorithmNames }
             serverAlgorithms = peerKeyExchangeAlgorithms
             clientHostKeyAlgorithms = self.supportedHostKeyAlgorithms
             serverHostKeyAlgorithms = peerHostKeyAlgorithms
         case .server:
             clientAlgorithms = peerKeyExchangeAlgorithms
-            serverAlgorithms = Self.supportedKeyExchangeAlgorithms
+            serverAlgorithms = role.keyExchangeAlgorithms.flatMap { $0.keyExchangeAlgorithmNames }
```
Lukasa, line 386:
```suggestion
            serverAlgorithms = role.keyExchangeAlgorithmNames
```

Sources/NIOSSH/Key Exchange/SSHKeyExchangeStateMachine.swift
```diff
@@ -369,13 +377,13 @@ struct SSHKeyExchangeStateMachine {
 
         switch self.role {
         case .client:
-            clientAlgorithms = Self.supportedKeyExchangeAlgorithms
+            clientAlgorithms = role.keyExchangeAlgorithms.flatMap { $0.keyExchangeAlgorithmNames }
```
Lukasa, line 380:
```suggestion
            clientAlgorithms = role.keyExchangeAlgorithmNames
```

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -175,16 +187,100 @@ extension NIOSSHPublicKey {
...
+    customAlgorithmsLock.lock()
+    defer { customAlgorithmsLock.unlock() }
+    return _customKeyExchangeAlgorithms
+}
+
+internal let customAlgorithmsLock = NSLock()
+fileprivate var _customTransportProtectionSchemes = [NIOSSHTransportProtection.Type]()
+fileprivate var _customKeyExchangeAlgorithms = [NIOSSHKeyExchangeAlgorithmProtocol.Type]()
+fileprivate var _customPublicKeyAlgorithms: [NIOSSHPublicKeyProtocol.Type] = []
+fileprivate var _customSignatures: [NIOSSHSignatureProtocol.Type] = []
```
Lukasa, line 282: Can we nest these inside an `enum` without cases just to give them a namespace?

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -175,16 +187,100 @@ extension NIOSSHPublicKey {
...
+    return _customTransportProtectionSchemes
+}
+
+internal var customKeyExchangeAlgorithms: [NIOSSHKeyExchangeAlgorithmProtocol.Type] {
+    customAlgorithmsLock.lock()
+    defer { customAlgorithmsLock.unlock() }
+    return _customKeyExchangeAlgorithms
+}
+
+internal let customAlgorithmsLock = NSLock()
```
Lukasa, line 278: We should probably use fine-grained locking here, so can we have one lock per field? It'll reduce contention on the locks.

Lukasa: Additionally, instead of `NSLock` can we add a dependency on `NIOConcurrencyHelpers` and use that `Lock`?

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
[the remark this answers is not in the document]
Lukasa: Can we remove `writeWithoutHeader` then? It'd be good to avoid public API surface we're not sure we need.

Sources/NIOSSH/Role.swift
```diff
@@ -34,4 +34,26 @@ public enum SSHConnectionRole {
...
+        switch self {
+        case .client(let client):
+            return client.transportProtectionSchemes
+        case .server(let server):
+            return server.transportProtectionSchemes
+        }
+    }
+    
+    internal var keyExchangeAlgorithmNames: [Substring] {
+        keyExchangeAlgorithms.flatMap { $0.keyExchangeAlgorithmNames }
```
Lukasa, line 48: Nit:

```suggestion
        self.keyExchangeAlgorithms.flatMap { $0.keyExchangeAlgorithmNames }
```

Sources/NIOSSH/TransportProtection/AESGCM.swift
```diff
@@ -43,7 +43,7 @@ internal class AESGCMTransportProtection {
     required init(initialKeys: NIOSSHSessionKeys) throws {
         guard initialKeys.outboundEncryptionKey.bitCount == Self.keySizes.encryptionKeySize * 8,
             initialKeys.inboundEncryptionKey.bitCount == Self.keySizes.encryptionKeySize * 8 else {
-            throw NIOSSHError.invalidKeySize
+N            throw NIOSSHError.invalidKeySize
```
Lukasa, line 46: Nit:

```suggestion
             throw NIOSSHError.invalidKeySize
```

Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
```diff
@@ -31,34 +31,34 @@ public struct SSHChildChannelOptions {
     public static let peerMaximumMessageLength: SSHChildChannelOptions.Types.PeerMaximumMessageLengthOption = .init()
 }
 
-extension SSHChildChannelOptions {
-    public enum Types {}
+public extension SSHChildChannelOptions {
```
Lukasa, line 34: NIO style is to place the access modifiers within the extensions, not on the extension itself.

Sources/NIOSSH/Child Channels/ChildChannelOptions.swift
```diff
@@ -31,34 +31,34 @@ public struct SSHChildChannelOptions {
...
+    enum Types {}
 }
 
-extension SSHChildChannelOptions.Types {
+public extension SSHChildChannelOptions.Types {
```
Lukasa, line 38: NIO style is to place the access modifiers within the extensions, not on the extension itself.

Sources/NIOSSH/Child Channels/ChildChannelStateMachine.swift
```diff
@@ -24,8 +24,8 @@ struct ChildChannelStateMachine {
     }
 }
 
-extension ChildChannelStateMachine {
-    fileprivate enum State: Hashable {
+private extension ChildChannelStateMachine {
```
Lukasa, line 27: NIO style is to place the access modifiers within the extensions, not on the extension itself.

Sources/NIOSSH/Child Channels/ChildChannelUserEvents.swift
```diff
@@ -299,7 +299,7 @@ extension SSHChannelRequestEvent {
     /// Constructs a channel request event and wraps it up in an Any.
     ///
     /// This is usually used just prior to firing this into the pipeline.
-    internal static func fromMessage(_ message: SSHMessage.ChannelRequestMessage) -> Any? {
+    static func fromMessage(_ message: SSHMessage.ChannelRequestMessage) -> Any? {
```
Lukasa, line 302: This change is unnecessary, please remove.

Lukasa (changes requested):
I started reviewing this, but the change is _massive_ because it has reformatted the codebase. Can you please use the exact version of swift-format that CI is using and re-run to clean up the format commit? [Our Dockerfile](https://github.com/apple/swift-nio-ssh/blob/main/docker/Dockerfile) can assist with that.

gwynne:
@Lukasa Rebased out all the `swiftformat` changes and re-applied the correct version, diff should be much better now 🤞

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -175,16 +188,107 @@ extension NIOSSHPublicKey {
...
+    }
+
+    internal static var customSignatures: [NIOSSHSignatureProtocol.Type] {
+        _CustomAlgorithms.signaturesLock.withLock {
+            _CustomAlgorithms.signatures
+        }
+    }
+}
+
+public enum NIOSSHAlgorithms {
```
Lukasa, line 219: I remain uncertain about doing this with global state. Can we consider placing this on per connection configuration instead of using this global config?

gwynne: @Joannis Can you weigh in on this?

Joannis: I agree that a per-connection basis is a better solution. But we will need a registry for sure, since we can have a variety of Host Keys for clients, and accepted public keys for servers

JaapWijnen: @Lukasa @gwynne While I was picking this up I was discussing this with Joannis and want to introduce a per connection solution. But while discussing with the way the custom keys currently work we could get rid of the BackingKey enums that are currently internal in the library. This would eliminate a lot of switch statements and simplify a lot of methods quite a bit. We don't immediately see any issues, but I wanted to check in with you, what are your thoughts on this?

Lukasa: I don't think the switch statements are that big a deal: they provide fast-paths for the most common use-cases where the compiler can be aware of exactly what the implementation is. The extensible backing pattern should certainly run this through a protocol though.

JaapWijnen: That's true it would make the whole implementation more consistent however and remove a lot of boilerplate/duplication. So not something the user benefits from perse but still could be interesting in terms of maintainability/accessible for new devs checking out the repo? I put in quite some work to get it towards that end. Would you be interested in seeing a PR? (it still needs a little work but if you're not interested in merging that kind of work in to the repo I'll probably leave it as is and just finish up your last remarks regarding not using global state for registering algorithms, which btw already needs quite a bit of moving around of methods in order to pass the registered algorithms down to the right ByteBuffer extensions) I'm hoping to finish up the work and then ideally divide it into a few PR's so it's a bit more manageable for you to review. Anyway let me know! :)

Joannis:
@Lukasa please continue on the new track

PR #75, opened 2021-02-19
Allows clients tolerate lines before version

### LINKED ISSUE #74: Clients should tolerate arbitrary lines before SSH version in banner
From https://tools.ietf.org/html/rfc4253#section-4.2:

> The server MAY send other lines of data before sending the version
> string.  Each line SHOULD be terminated by a Carriage Return and Line
> Feed.  Such lines MUST NOT begin with "SSH-", and SHOULD be encoded
> in ISO-10646 UTF-8 [RFC3629] (language is not specified).

artemredkin:
Motivation:
RFC 4253, section 4.2 allows for server to send some lines before
version in version exchange. Right now we expect that version start with
`SSH-` only. We should support this.

Modifications:
 - Modifies version verification to allow clients tolerate lines before
   version
 - Adds tests that verify that client accepts lines and server doesn't

Result:
Closes #74

### DIFF
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
@@ -15,18 +15,71 @@
 protocol AcceptsVersionMessages {}
 
 extension AcceptsVersionMessages {
-    func receiveVersionMessage(_ version: String) throws {
-        try self.validateVersion(version)
+    func receiveVersionMessage(_ banner: String, role: SSHConnectionRole) throws {
+        try self.validateBanner(banner, role: role)
     }
 
-    private func validateVersion(_ version: String) throws {
-        guard version.count > 7, version.hasPrefix("SSH-") else {
-            throw NIOSSHError.unsupportedVersion(version)
+    // From RFC 4253:
+    //
+    // > Protocol Version Exchange
+    // >
+    // > When the connection has been established, both sides MUST send an
+    // > identification string.  This identification string MUST be
+    // >   SSH-protoversion-softwareversion SP comments CR LF
+    // > Since the protocol being defined in this set of documents is version
+    // > 2.0, the 'protoversion' MUST be "2.0".  The 'comments' string is
+    // > OPTIONAL.  If the 'comments' string is included, a 'space' character
+    // > (denoted above as SP, ASCII 32) MUST separate the 'softwareversion'
+    // > and 'comments' strings.  The identification MUST be terminated by a
+    // > single Carriage Return (CR) and a single Line Feed (LF) character
+    // > (ASCII 13 and 10, respectively).  Implementers who wish to maintain
+    // > compatibility with older, undocumented versions of this protocol may
+    // > want to process the identification string without expecting the
+    // > presence of the carriage return character for reasons described in
+    // > Section 5 of this document.  The null character MUST NOT be sent.
+    // > The maximum length of the string is 255 characters, including the
+    // > Carriage Return and Line Feed.
+    // >
+    // > The part of the identification string preceding the Carriage Return
+    // > and Line Feed is used in the Diffie-Hellman key exchange (see Section
+    // > 8).
+    // >
+    // > The server MAY send other lines of data before sending the version
+    // > string.  Each line SHOULD be terminated by a Carriage Return and Line
+    // > Feed.  Such lines MUST NOT begin with "SSH-", and SHOULD be encoded
+    // > in ISO-10646 UTF-8 [RFC3629] (language is not specified).  Clients
+    // > MUST be able to process such lines.  Such lines MAY be silently
+    // > ignored, or MAY be displayed to the client user.  If they are
+    // > displayed, control character filtering, as discussed in [SSH-ARCH],
+    // > SHOULD be used.  The primary use of this feature is to allow TCP-
+    // > wrappers to display an error message before disconnecting
+    private func validateBanner(_ banner: String, role: SSHConnectionRole) throws {
+        switch role {
+        case .client:
+            // split by \n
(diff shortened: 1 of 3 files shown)


Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
```diff
@@ -15,18 +15,69 @@
...
+                throw NIOSSHError.protocolViolation(protocolName: "version exchange", violation: "version string not found")
+            }
         }
-        let start = version.index(version.startIndex, offsetBy: 4)
-        let end = version.index(start, offsetBy: 3)
-        guard version[start ..< end] == "2.0" else {
-            throw NIOSSHError.unsupportedVersion(version)
+    }
+
+    private func validateVersion<S: StringProtocol>(_ version: S) throws -> Bool {
```
Lukasa, line 72: There's no real reason to make this generic over `StringProtocol`: we're only gonna operate on `Substring` so let's just accept that.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
```diff
@@ -15,18 +15,69 @@
...
+        if version.hasPrefix("SSH-") {
+            let start = version.index(version.startIndex, offsetBy: 4)
+            let end = version.index(start, offsetBy: 3)
```
Lukasa, line 75: These indices need to be validated: if the string was `"SSH-"` then both would be ineligible to be subscripted and the slice would be invalid. We need to confirm that the total count is appropriate.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
```diff
@@ -15,18 +15,69 @@
...
+            }
         }
-        let start = version.index(version.startIndex, offsetBy: 4)
-        let end = version.index(start, offsetBy: 3)
-        guard version[start ..< end] == "2.0" else {
-            throw NIOSSHError.unsupportedVersion(version)
+    }
+
+    private func validateVersion<S: StringProtocol>(_ version: S) throws -> Bool {
+        if version.hasPrefix("SSH-") {
```
Lukasa, line 73: Also all of this is being done on the grapheme clusters. We should consider doing all the work on the `String` `UTF8View` instead, which will give us much better performance as we don't need to worry about unicode here.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
```diff
@@ -15,18 +15,69 @@
...
+    // > wrappers to display an error message before disconnecting
+    private func validateBanner(_ banner: String, role: SSHConnectionRole) throws {
+        switch role {
+        case .client:
+            for line in banner.split(separator: "\n") {
+                if try self.validateVersion(line) {
+                    return
+                }
+            }
+            throw NIOSSHError.protocolViolation(protocolName: "version exchange", violation: "version string not found")
```
Lukasa, line 64: I don't think we have a test for this error, do we?

Lukasa (changes requested):
Cool, great start @artemredkin! I've got a few notes.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
artemredkin: we call this function with both String _and_ Substring, we call it with string on line 66. Maybe alternative is to covert string to subtring there?

Lukasa: Oh yeah, let's do that. `String` -> `Substring` conversion is free.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
artemredkin: good catch, yes, added a test

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
artemredkin: done

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
artemredkin: right, fixed

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
artemredkin: great idea, done!

Lukasa (changes requested):
One minor nit.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsVersionMessages.swift
```diff
@@ -15,18 +15,70 @@
...
+    // > MUST be able to process such lines.  Such lines MAY be silently
+    // > ignored, or MAY be displayed to the client user.  If they are
+    // > displayed, control character filtering, as discussed in [SSH-ARCH],
+    // > SHOULD be used.  The primary use of this feature is to allow TCP-
+    // > wrappers to display an error message before disconnecting
+    private func validateBanner(_ banner: String, role: SSHConnectionRole) throws {
+        switch role {
+        case .client:
+            // split by \n
+            for line in banner.utf8.split(separator: 10) {
```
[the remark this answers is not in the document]
artemredkin: ah, looks nice, fixed

Lukasa (approved):
Nice, LGTM.

PR #76, opened 2021-03-01
Adds emittion of user authentication success event.

### LINKED ISSUE #39: Introduce `handshake complete/failed` events
Sometimes its better to be sure that main parent channel is established without errors, before child channels are created

artemredkin:
Motivation:
Currently there is now way to wait for successful SSH session
establishment, the only way to continue is to try to create a child
channel. Given that most clients are expecting the session to be active
when both key exchange and user authenticaiton complete, I propose to
emit UserAuthSuccessEvent.  That way client can wait for that event or
fail if there is an error in the pipeline.

Modifications:
 - Adds UserAuthSuccessEvent event
 - Adds .event case to StateMachineInboundProcessResult
 - Adds emittion of event if StateMachineInboundProcessResult is .event
 - Adds two tests

Result:
Closes #39

### DIFF
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -16,6 +16,11 @@ protocol AcceptsUserAuthMessages {
     var userAuthStateMachine: UserAuthenticationStateMachine { get set }
 }
 
+/// This event indicates that server accepted our response to authentication challenge. SSH session can be considered active after that.
+public struct UserAuthSuccessEvent: Hashable {
+    public init() {}
+}
+
 extension AcceptsUserAuthMessages {
     mutating func receiveServiceRequest(_ message: SSHMessage.ServiceRequestMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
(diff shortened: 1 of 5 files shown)


artemredkin:

artemredkin:

tomerd:

tomerd:

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
```diff
@@ -16,6 +16,9 @@ protocol AcceptsUserAuthMessages {
     var userAuthStateMachine: UserAuthenticationStateMachine { get set }
 }
 
+/// This event indicates that server accepted our response to authentication challenge. SSH session can be considered active after that.
+public struct UserAuthSuccessEvent {}
```
Lukasa, line 20: Can we add a `public init` to this? Otherwise users can't construct instances of it. We should also make it `Hashable`.

Lukasa (approved):
Nice clean patch, I really like it.

Lukasa:
One little nit and then we're good to go.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
artemredkin: done!

PR #80, opened 2021-03-16
Add support from synchronous options

glbrntt:
Motivation:

NIO 2.27.0 added a customisation point to `Channel` to allow options to
be fetched synchronously if the caller is on the appropriate event loop.

Modifications:

- Add synchronous options to SSHChildChannel
- Promote event loop assertions to preconditions

Result:

Callers can fetch channel options synchronously on 'SSHChildChannel's.


weissi (approved):
Nice! lgtm

Lukasa (approved):

PR #83, opened 2021-05-11
Add a benchmark for SSH throughput.

Lukasa:
Motivation:

We should have more benchmarks.

Modifications:

Add a throughput benchmark that just sends a bunch of data through.

Result:

We'll have more data.


Lukasa:
@swift-server-bot test perf please

PR #84, opened 2021-05-11
Use the ContiguousBytes functions for setting data

Lukasa:
Motivation:

We should use setContiguousBytes instead of setBytes to avoid hitting
the slow path.

Modifications:

- Use setContiguousBytes.

Result:

Speedy speedy


weissi (approved):
👍

PR #90, opened 2021-08-24
fixes compilation warning

artemredkin:
Motivation:
Method Channel.removeHandlers(channel: Channel) was deprecated in 2.32,
we should use recommended method instead.

Modifications:
 - Bump swift-nio version to 2.32 to ensue that new methods are
   available
 - Use new method to remove all handlers


glbrntt (approved):
agh, thanks Artem!

PR #91, opened 2021-09-22
Update to Swift 5.5 release in CI

Lukasa:


fabianfett:
We seem to have the jazzy problem here as well.

fabianfett (approved):

PR #92, opened 2021-09-23
Fix doc generation and jazzy version

Lukasa:
Motivation:

The combination of Ruby and Jazzy are throwing up issues in CI on older
Ubuntu versions.

Modifications:

- Only install Jazzy on focal
- Fix doc generation script to use more recent source-kitten syntax

Results:

- CI doesn't complain
- Docs get built

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update && apt-get install -y lsof dnsutils netcat-openbsd net-tools 
 # ruby and jazzy for docs generation, only on focal
 RUN apt-get update && apt-get install -y ruby ruby-dev libsqlite3-dev build-essential
 RUN if [ "${ubuntu_version}" == "focal" ] ; then echo "gem: --no-document" > ~/.gemrc; fi
- RUN if [ "${ubuntu_version}" == "focal" ] ; then  gem install jazzy; fi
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -16,12 +16,10 @@ ENV LANGUAGE en_US.UTF-8
...
 
-# ruby and jazzy for docs generation
+# ruby and jazzy for docs generation, only on focal
 RUN apt-get update && apt-get install -y ruby ruby-dev libsqlite3-dev build-essential
-# switch off gem docs building
-RUN echo "gem: --no-document" > ~/.gemrc
-# jazzy no longer works on xenial as ruby is too old.
-RUN if [ "${ubuntu_version}" != "xenial" ] ; then  gem install jazzy; fi
+RUN if [ "${ubuntu_version}" == "focal" ] ; then echo "gem: --no-document" > ~/.gemrc; fi
+ RUN if [ "${ubuntu_version}" == "focal" ] ; then  gem install jazzy; fi
```
fabianfett, line 22:
```suggestion
RUN if [ "${ubuntu_version}" == "focal" ] ; then  gem install jazzy; fi
```

fabianfett (approved):
Small typo. Please fix before merge.

PR #95, opened 2021-09-27
Fix crash sending window increment after close

Lukasa:
Motivation:

When reads have yet to be delivered to a channel and we receive a
CHANNEL_CLOSE message, we forcibly deliver all of the reads. As part of
doing this, we process the window size changes and potentially send
window increment messages. These messages are invalid to send at this
time, and the state machine rightly forbids us from sending them, which
causes a crash.

Modifications:

- Change the code to only futz with the channel window if the channel is
  still open.

Result:

No crashing after channel close.


fabianfett (approved):
LGTM

Issue #50, opened 2020-11-20
Third-Party Cryptographic Implementations


Joannis:
For implementing [RSA support](https://github.com/Joannis/swift-nio-ssh-1/pull/1/files), which will not be landing in the main library as discussed, we'll need to support external implementations for at least public-private keys. I'd also recommend looking into the same for the symmetric encryption as seen in #48

Lukasa:
I think a good idea here is to try to throw together a few implementations and see what the common API surface needs to be, and then to offer that as an extension point. We have some time to workshop APIs here.

Lukasa:
@gwynne this seems like something you'd find fun!

PR #97, opened 2021-10-13
Adds support for custom transport protection algorithms

artemredkin:
Motivation:
In some cases AES GCM might not be supported by a server/client, and
clients might require custom algorithms to be implemented in their
projects.

Modifications:
 - Makes NIOSSHTransportProtection, NIOSSHSessionKeys and ExpectedKeySizes public
 - Makes transport algorithms list configurable
 - Adds sequence number of incoming and outgoing messages to parser and
   serializer
 - Tests

### DIFF
--- a/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
+++ b/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
@@ -60,11 +60,7 @@ struct SSHConnectionStateMachine {
     /// The state of this state machine.
     private var state: State
 
-    private static let defaultTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
-        AES256GCMOpenSSHTransportProtection.self, AES128GCMOpenSSHTransportProtection.self,
-    ]
-
-    init(role: SSHConnectionRole, protectionSchemes: [NIOSSHTransportProtection.Type] = Self.defaultTransportProtectionSchemes) {
+    init(role: SSHConnectionRole, protectionSchemes: [NIOSSHTransportProtection.Type] = Constants.bundledTransportProtectionSchemes) {
         self.state = .idle(IdleState(role: role, protectionSchemes: protectionSchemes))
     }
 
--- a/Sources/NIOSSH/Constants.swift
+++ b/Sources/NIOSSH/Constants.swift
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-enum Constants {
+public enum Constants {
     static let version = "SSH-2.0-SwiftNIOSSH_1.0"
+
+    public static let bundledTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
+        AES256GCMOpenSSHTransportProtection.self, AES128GCMOpenSSHTransportProtection.self,
+    ]
 }
--- a/Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
+++ b/Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
@@ -46,18 +46,27 @@ extension KeyExchangeResult: Equatable {}
 /// Of these types, the encryption keys and the MAC keys are intended to be secret, and so
 /// we store them in the `SymmetricKey` types. The IVs do not need to be secret, and so are
 /// stored in regular heap buffers.
-struct NIOSSHSessionKeys {
-    var initialInboundIV: [UInt8]
+public struct NIOSSHSessionKeys {
+    public var initialInboundIV: [UInt8]
 
-    var initialOutboundIV: [UInt8]
+    public var initialOutboundIV: [UInt8]
 
-    var inboundEncryptionKey: SymmetricKey
+    public var inboundEncryptionKey: SymmetricKey
 
-    var outboundEncryptionKey: SymmetricKey
+    public var outboundEncryptionKey: SymmetricKey
 
-    var inboundMACKey: SymmetricKey
+    public var inboundMACKey: SymmetricKey
 
-    var outboundMACKey: SymmetricKey
+    public var outboundMACKey: SymmetricKey
+
+    public init(initialInboundIV: [UInt8], initialOutboundIV: [UInt8], inboundEncryptionKey: SymmetricKey, outboundEncryptionKey: SymmetricKey, inboundMACKey: SymmetricKey, outboundMACKey: SymmetricKey) {
+        self.initialInboundIV = initialInboundIV
+        self.initialOutboundIV = initialOutboundIV
+        self.inboundEncryptionKey = inboundEncryptionKey
+        self.outboundEncryptionKey = outboundEncryptionKey
+        self.inboundMACKey = inboundMACKey
+        self.outboundMACKey = outboundMACKey
+    }
 }
 
 extension NIOSSHSessionKeys: Equatable {}
@@ -68,10 +77,18 @@ extension NIOSSHSessionKeys: Equatable {}
 /// hash function invocations. The output of these hash functions is truncated to an appropriate
 /// length as needed, which means we need to ensure the code doing the calculation knows how
 /// to truncate appropriately.
-struct ExpectedKeySizes {
-    var ivSize: Int
+public struct ExpectedKeySizes {
+    public var ivSize: Int
 
-    var encryptionKeySize: Int
+    public var encryptionKeySize: Int
 
-    var macKeySize: Int
+    public var macKeySize: Int
+
+    public init(ivSize: Int, encryptionKeySize: Int, macKeySize: Int) {
+        self.ivSize = ivSize
+        self.encryptionKeySize = encryptionKeySize
+        self.macKeySize = macKeySize
+    }
 }
+
+extension ExpectedKeySizes: Hashable {}
--- a/Sources/NIOSSH/Role.swift
+++ b/Sources/NIOSSH/Role.swift
@@ -37,4 +37,13 @@ public enum SSHConnectionRole {
             return true
         }
     }
+
+    internal var transportProtectionSchemes: [NIOSSHTransportProtection.Type] {
+        switch self {
+        case .client(let configuration):
+            return configuration.transportProtectionSchemes
+        case .server(let configuration):
+            return configuration.transportProtectionSchemes
+        }
+    }
 }
--- a/Sources/NIOSSH/SSHClientConfiguration.swift
+++ b/Sources/NIOSSH/SSHClientConfiguration.swift
@@ -23,11 +23,22 @@ public struct SSHClientConfiguration {
     /// The global request delegate to be used with this client.
     public var globalRequestDelegate: GlobalRequestDelegate
 
+    /// Supported data encryption algorithms
+    public var transportProtectionSchemes: [NIOSSHTransportProtection.Type]
+
     public init(userAuthDelegate: NIOSSHClientUserAuthenticationDelegate,
                 serverAuthDelegate: NIOSSHClientServerAuthenticationDelegate,
                 globalRequestDelegate: GlobalRequestDelegate? = nil) {
+        self.init(userAuthDelegate: userAuthDelegate, serverAuthDelegate: serverAuthDelegate, globalRequestDelegate: globalRequestDelegate, transportProtectionSchemes: Constants.bundledTransportProtectionSchemes)
+    }
+
+    public init(userAuthDelegate: NIOSSHClientUserAuthenticationDelegate,
+                serverAuthDelegate: NIOSSHClientServerAuthenticationDelegate,
+                globalRequestDelegate: GlobalRequestDelegate? = nil,
+                transportProtectionSchemes: [NIOSSHTransportProtection.Type]) {
         self.userAuthDelegate = userAuthDelegate
         self.serverAuthDelegate = serverAuthDelegate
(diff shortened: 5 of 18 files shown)


Sources/NIOSSH/Constants.swift
```diff
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-enum Constants {
+public enum Constants {
```
artemredkin, line 15: Not sure this is the best place, maybe we need to type `Defaults` or smth?

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
```diff
@@ -152,6 +149,7 @@ struct SSHConnectionStateMachine {
             }
         case .keyExchange(var state):
             guard let message = try state.parser.nextPacket() else {
+                self.state = .keyExchange(state)
```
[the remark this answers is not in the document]
artemredkin: ok, so far I caught it only in `.active` state and it breaks if we do not receive a full packet, only length, in parser we set state to `.waitingForBytes(length)` but due to CoW `SSHConnectionStateMachine` still has an old value of parser. And it's not safe to unwind the buffer as well, since `decryptFirstBlock` might be stateful as well. Now I need to figure out a test for this

Joannis (approved):
Looks like a solid PR to me. I don't have anything to add to the code, but I don't think #50 can be closed after this PR. This PR only covers transport protection.

artemredkin:
Agreed, but at least it should be referenced there, thank you!

Sources/NIOSSH/Constants.swift
Lukasa: I agree with you, I think `Defaults` would be better.

Sources/NIOSSH/Constants.swift
```diff
@@ -12,6 +12,10 @@
...
     static let version = "SSH-2.0-SwiftNIOSSH_1.0"
+
+    public static let defaultTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
```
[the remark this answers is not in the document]
Lukasa: I'm not sure that there's a better alternative than this.

Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
```diff
@@ -46,18 +46,18 @@ extension KeyExchangeResult: Equatable {}
 /// Of these types, the encryption keys and the MAC keys are intended to be secret, and so
 /// we store them in the `SymmetricKey` types. The IVs do not need to be secret, and so are
 /// stored in regular heap buffers.
-struct NIOSSHSessionKeys {
-    var initialInboundIV: [UInt8]
+public struct NIOSSHSessionKeys {
```
Lukasa, line 49: Can we add a `public init` as well? There's no good reason to prevent users from constructing one of these.

Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
```diff
@@ -68,10 +68,16 @@ extension NIOSSHSessionKeys: Equatable {}
 /// hash function invocations. The output of these hash functions is truncated to an appropriate
 /// length as needed, which means we need to ensure the code doing the calculation knows how
 /// to truncate appropriately.
-struct ExpectedKeySizes {
-    var ivSize: Int
+public struct ExpectedKeySizes {
```
Lukasa, line 71: This may as well be `Hashable`.

Sources/NIOSSH/Constants.swift
artemredkin: Decided to change this a bit, I will rename constant to `bundledTransportProtectionSchemes`, I quite like this idea from [Joannis](https://github.com/Joannis)'s PR, wdyt?

Sources/NIOSSH/SSHPacketSerializer.swift
```diff
@@ -22,6 +22,15 @@ struct SSHPacketSerializer {
...
     private var state: State = .initialized
+    private(set) var sequence: UInt32 = 0
+
+    private mutating func increment() {
+        if self.sequence == UInt32.max {
+            self.sequence = 0
+        } else {
+            self.sequence += 1
+        }
+    }
```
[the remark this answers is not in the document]
artemredkin: done

Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
artemredkin: done

Sources/NIOSSH/Key Exchange/SSHKeyExchangeResult.swift
artemredkin: done

Sources/NIOSSH/SSHClientConfiguration.swift
```diff
@@ -23,11 +23,22 @@ public struct SSHClientConfiguration {
     /// The global request delegate to be used with this client.
     public var globalRequestDelegate: GlobalRequestDelegate
 
+    /// Supported data encryption algorithms
+    public var transportProtectionSchemes: [NIOSSHTransportProtection.Type]
```
artemredkin, line 27: I'm suggesting to put all configurable cryptographic algorithms in `SSHClientConfiguration` and `SSHServerConfiguration`, next candidate could be `keyExchangeAlgorithms`. And all static members will become `bundled*`. Wdyt? cc @Joannis

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
```diff
@@ -60,11 +60,11 @@ struct SSHConnectionStateMachine {
     /// The state of this state machine.
     private var state: State
 
-    private static let defaultTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
+    public static let bundledTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
```
artemredkin, line 63: I decided to make this public so clients can have an ability to specify a fully custom set of algorithms (i.e. exclude bundled). But if they need to add custom, they will be able to specify `bundled` + custom

artemredkin: Ah, `NIOSSHConnectionStateMachine` is not public, moved to `SSHConnectionRole`

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
```diff
@@ -87,10 +87,10 @@ protocol NIOSSHTransportProtection: AnyObject {
     /// length, the padding, or the MAC), and update source to indicate the consumed bytes.
     /// It must also perform any integrity checking that
     /// is required and throw if the integrity check fails.
-    func decryptAndVerifyRemainingPacket(_ source: inout ByteBuffer) throws -> ByteBuffer
+    func decryptAndVerifyRemainingPacket(_ source: inout ByteBuffer, sequenceNumber: UInt32) throws -> ByteBuffer
 
     /// Encrypt an entire outbound packet
-    func encryptPacket(_ packet: NIOSSHEncryptablePayload, to outboundBuffer: inout ByteBuffer) throws
+    func encryptPacket(_ packet: NIOSSHEncryptablePayload, sequenceNumber: UInt32, to outboundBuffer: inout ByteBuffer) throws
```
artemredkin, line 93: One thing I haven't solved yet, is how to use `NIOSSHEncryptablePayload` to write tests for custom transport implementations. It uses internal type (SSHMessage), so there is no way to initialize it. I see three options:
1. Make it a protocol
2. Make it an enum (`SSHMessage|ByteBuffer)
3. Make `encryptPacket` similar to decrypt, where we will be passing an `inout` buffer with message already written there expecting client to encrypt it inplace
wdyt?

artemredkin: (right now I'm in favour of 3 tbh)

artemredkin: I think many implementation will begin with:
```
        let packetLengthIndex = outboundBuffer.writerIndex
        let packetLengthLength = MemoryLayout<UInt32>.size
        let packetPaddingIndex = outboundBuffer.writerIndex + packetLengthLength
        let packetPaddingLength = MemoryLayout<UInt8>.size

        outboundBuffer.moveWriterIndex(forwardBy: packetLengthLength + packetPaddingLength)
```
maybe we should abstract in serializer?

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
```diff
@@ -87,10 +87,10 @@ protocol NIOSSHTransportProtection: AnyObject {
     /// length, the padding, or the MAC), and update source to indicate the consumed bytes.
     /// It must also perform any integrity checking that
     /// is required and throw if the integrity check fails.
-    func decryptAndVerifyRemainingPacket(_ source: inout ByteBuffer) throws -> ByteBuffer
+    func decryptAndVerifyRemainingPacket(_ source: inout ByteBuffer, sequenceNumber: UInt32) throws -> ByteBuffer
```
artemredkin, line 90: Actually, two additional questions before we make this API public:
1. Why do we return a `ByteBuffer` here instead of in-place mutation?
2. There is a bit of a lost symmetry here, in plaintext mode padding and packet length are consumed by parser, but in ciphertext mode we have to do it manually, should this be handled by parser instead? And decryptors should only decrypt data in place and leave it? (Maybe we should drop mac bytes here?)

artemredkin: I think parser can handle mac bytes, we have `macBytes` property, we can use it to remove last bytes from the packet

Sources/NIOSSH/Role.swift
```diff
@@ -14,6 +14,10 @@
 
 /// The role of a given party in an SSH connection.
 public enum SSHConnectionRole {
+    public static let bundledTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
```
Lukasa, line 17: I don't think this makes sense on `SSHConnectionRole`. This should go on config objects. We can avoid some duplication by using an internal central location.

Sources/NIOSSH/SSHClientConfiguration.swift
Lukasa: Works for me.

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
Lukasa: 1. As the doc comment says: it is expected to update the buffer to indicate what it consumed, and return the decrypted buffer. We avoid in-place mutation because we do not want to guarantee that all encryption algorithms have the exact same size input and output.
2. The absence of symmetry is because in the input case we need to be able to decrypt just the length so we know how much further data to read, and we wanted to minimise the number of times we need to do that operation. I don't think the parser _can_ do it for us because the actual length is affected by the protection scheme.

We could pass the MAC separately, but I'm not sure we gain much do we?

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
Lukasa: I suspect (3) is the right approach, yes. We can also abstract the write pattern in the serializer as well. Can I suggest that we do that as a separate PR though to avoid this one growing further?

artemredkin: agreed 👍

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
artemredkin: let me think about this further, I'll revisit this together with refactoring `encryptPacket`

Sources/NIOSSH/Role.swift
artemredkin: moved to `Constants`

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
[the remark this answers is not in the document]
artemredkin: This was fixed in #102

Sources/NIOSSH/Constants.swift
Lukasa: LGTM.

Sources/NIOSSH/TransportProtection/SSHTransportProtection.swift
Lukasa: Just for my clarification, did this work get done?

artemredkin: I opened a draft here:
https://github.com/artemredkin/swift-nio-ssh/pull/1
but so far I only refactored `encryptPacket`, we probably should have two separate PRs for `encryptPacket` and `decryptAndVerifyRemainingPacket`

Lukasa: Ok great, so you'd like to have this PR merged without further changes to this and we'll revisit in a follow-on PR?

artemredkin: yes! otherwise it's already quite large

PR #98, opened 2021-11-09
Feature/userauth banners

DominikHorn:
Hi everyone,

Upon using this excellent package to build an SSH client, I stumbled upon the following issue:

SSH servers may ask clients to display a userauth banner message during the user authentication step of the protocol. This mechanism is documented in RFC 4252. Since the underlying SSH_MSG_USERAUTH_BANNER message has thus far been entirely unimplemented, receiving userauth banners results in an 'unknownType(53)' error. This makes it impossible to connect to servers that send SSH_MSG_USERAUTH_BANNER using clients built with swift-nio-ssh - at least as far as I know.

This pull request aims to fix said issue, namely by implementing the userauth banner spec from RFC 4252.

I've thus far only encountered userauth banners that contained information about the underlying SSH server's hardware. I'd argue that this information is probably protection worthy, i.e., should not be disclosed to unauthorized third parties. I've thus taken the liberty to send banner messages from swift-nio-ssh servers right before user auth succeeds, to ensure that only authorized clients will receive banner message contents. This should technically be standards compliant, as RFC 4252 states: "_The SSH server may send an SSH_MSG_USERAUTH_BANNER message at any time after this authentication protocol starts and before authentication is successful._".
However, I'm afraid this implementation choice rules out some use cases, as the one mentioned in the RFC: "_In some jurisdictions, sending a warning message before authentication may be relevant for getting legal protection. [...] This message contains text to be displayed to the client user before authentication is attempted._".
Right now, I can see two plausible options IMHO:

1. send SSH_MSG_USERAUTH_BANNER at some earlier point in time, namely, before the client attempts to authenticate
2. let package users choose not only if but also when the banner is to be sent. The API should in this case be very clear, e.g., by differentiating between PublicUserAuthBanner and SecretUserAuthBanner types that are automatically handled at the appropriate times instead of a single UserAuthBanner type.

I'd really like to hear your opinions on this topic to make sure that the userauth banner implementation is in the best interest of all stakeholders of this package. In general, I appreciate all the feedback you can give. 

Thank you very much :)

### DIFF
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -101,7 +101,7 @@ extension AcceptsUserAuthMessages {
         case .success:
             if let banner = banner {
                 // Send banner bundled with auth success to avoid leaking any information to unauthenticated clients.
-                // Note that this is by no means the only option
+                // Note that this is by no means the only option according to RFC 4252
                 return SSHMultiMessage(.userAuthBanner(.init(message: banner.message, languageTag: banner.languageTag)), .userAuthSuccess)
             }
 
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -52,7 +52,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveServiceAccept(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(self.transform(_:)))
+            return .possibleFutureMessage(future.map(Self.transform(_:)))
         } else {
             return .noMessage
         }
@@ -62,7 +62,12 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthRequest(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(self.transform(_:)))
+            var banner: SSHServerConfiguration.UserAuthBanner?
+            if case .server(let config) = role {
+                banner = config.banner
+            }
+
+            return .possibleFutureMessage(future.map({ Self.transform($0, banner: banner)}))
         } else {
             return .noMessage
         }
@@ -80,7 +85,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthFailure(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(self.transform(_:)))
+            return .possibleFutureMessage(future.map(Self.transform(_:)))
         } else {
             return .noMessage
         }
@@ -91,10 +96,10 @@ extension AcceptsUserAuthMessages {
       return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
     }
 
-    private func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
+    private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage, banner: SSHServerConfiguration.UserAuthBanner? = nil) -> SSHMultiMessage {
         switch result {
         case .success:
-            if case .server(let config) = self.role, let banner = config.banner {
+            if let banner = banner {
                 // Send banner bundled with auth success to avoid leaking any information to unauthenticated clients.
                 // Note that this is by no means the only option
                 return SSHMultiMessage(.userAuthBanner(.init(message: banner.message, languageTag: banner.languageTag)), .userAuthSuccess)
@@ -108,7 +113,7 @@ extension AcceptsUserAuthMessages {
         }
     }
 
-    private func transform(_ result: SSHMessage.UserAuthRequestMessage?) -> SSHMultiMessage? {
+    private static func transform(_ result: SSHMessage.UserAuthRequestMessage?) -> SSHMultiMessage? {
         result.map { SSHMultiMessage(.userAuthRequest($0)) }
     }
 }
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -19,9 +19,17 @@ protocol AcceptsUserAuthMessages {
 }
 
 /// This event indicates that server wants us to display the following message to the end user.
-public struct UserAuthBannerEvent: Hashable {
-    public let message: String
-    public let languageTag: String
+public struct NIOUserAuthBannerEvent: Hashable {
+    /// message to be displayed to end user
+    public var message: String
+
+    /// tag  identifying the language used for `message`, following RFC 3066
+    public var languageTag: String
+
+    public init(message: String, languageTag: String) {
+        self.message = message
+        self.languageTag = languageTag
+    }
 }
 
 /// This event indicates that server accepted our response to authentication challenge. SSH session can be considered active after that.
@@ -80,7 +88,7 @@ extension AcceptsUserAuthMessages {
 
     mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
       try self.userAuthStateMachine.receiveUserAuthBanner(message)
-      return .event(UserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
+      return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
     }
 
     private func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -67,7 +67,7 @@ extension AcceptsUserAuthMessages {
                 banner = config.banner
             }
 
-            return .possibleFutureMessage(future.map({ Self.transform($0, banner: banner)}))
+            return .possibleFutureMessage(future.map { Self.transform($0, banner: banner) })
         } else {
             return .noMessage
         }
@@ -92,8 +92,8 @@ extension AcceptsUserAuthMessages {
     }
 
     mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
-      try self.userAuthStateMachine.receiveUserAuthBanner(message)
-      return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
+        try self.userAuthStateMachine.receiveUserAuthBanner(message)
+        return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
     }
 
     private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage, banner: SSHServerConfiguration.UserAuthBanner? = nil) -> SSHMultiMessage {
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
@@ -14,6 +14,14 @@
 
 protocol AcceptsUserAuthMessages {
     var userAuthStateMachine: UserAuthenticationStateMachine { get set }
+
+    var role: SSHConnectionRole { get }
+}
+
+/// This event indicates that server wants us to display the following message to the end user.
+public struct UserAuthBannerEvent: Hashable {
+    public let message: String
+    public let languageTag: String
 }
 
 /// This event indicates that server accepted our response to authentication challenge. SSH session can be considered active after that.
@@ -36,7 +44,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveServiceAccept(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(Self.transform(_:)))
+            return .possibleFutureMessage(future.map(self.transform(_:)))
         } else {
             return .noMessage
         }
@@ -46,7 +54,7 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthRequest(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(Self.transform(_:)))
+            return .possibleFutureMessage(future.map(self.transform(_:)))
         } else {
             return .noMessage
         }
@@ -64,15 +72,26 @@ extension AcceptsUserAuthMessages {
         let result = try self.userAuthStateMachine.receiveUserAuthFailure(message)
 
         if let future = result {
-            return .possibleFutureMessage(future.map(Self.transform(_:)))
+            return .possibleFutureMessage(future.map(self.transform(_:)))
         } else {
             return .noMessage
         }
     }
 
-    private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
+    mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
+      try self.userAuthStateMachine.receiveUserAuthBanner(message)
+      return .event(UserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
+    }
+
+    private func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
         switch result {
         case .success:
+            if case .server(let config) = self.role, let banner = config.banner {
+                // Send banner bundled with auth success to avoid leaking any information to unauthenticated clients.
+                // Note that this is by no means the only option
+                return SSHMultiMessage(.userAuthBanner(.init(message: banner.message, languageTag: banner.languageTag)), .userAuthSuccess)
+            }
+
             return SSHMultiMessage(.userAuthSuccess)
         case .failure(let message):
             return SSHMultiMessage(.userAuthFailure(message))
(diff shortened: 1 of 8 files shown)


Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
```diff
@@ -14,6 +14,14 @@
 
 protocol AcceptsUserAuthMessages {
     var userAuthStateMachine: UserAuthenticationStateMachine { get set }
+
+    var role: SSHConnectionRole { get }
+}
+
+/// This event indicates that server wants us to display the following message to the end user.
+public struct UserAuthBannerEvent: Hashable {
```
Lukasa, line 22: For new types like this, we need to prefix them with the string `NIO`, to avoid type clashes.

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
```diff
@@ -14,6 +14,14 @@
...
+    public let message: String
+    public let languageTag: String
```
Lukasa, line 24: This is a bit of type design nitpicking, but for strict value types like this I would tend to recommend that we make both of these stored properties `var`. Can we also add a `public init` that allows users to construct one of these as well?

Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
```diff
@@ -64,15 +72,26 @@ extension AcceptsUserAuthMessages {
...
         }
     }
 
-    private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
+    mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
+      try self.userAuthStateMachine.receiveUserAuthBanner(message)
+      return .event(UserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
+    }
+
+    private func transform(_ result: NIOSSHUserAuthenticationResponseMessage) -> SSHMultiMessage {
```
Lukasa, line 86: Unfortunately we can't make this non-`static` without causing ourselves some trouble. NIO futures are "eager": if they have a value already and we attach a callback to them, that callback will be dispatched synchronously at the call site. Because the futures from which this is called are now closing over `self`, this risks violating the law of exclusivity at runtime: we're reading from `self` while engaged in a mutation of self.

Rather than make this a member function, can I propose we give it an extra argument that contains the optional `banner`? We can then close over that value as part of the callback.

Sources/NIOSSH/SSHMessages.swift
```diff
@@ -419,6 +431,11 @@ extension ByteBuffer {
                 return .userAuthFailure(message)
             case SSHMessage.UserAuthSuccessMessage.id:
                 return .userAuthSuccess
+            case SSHMessage.UserAuthBannerMessage.id:
+              guard let message = self.readUserAuthBannerMessage()  else {
+                return nil
+              }
+              return .userAuthBanner(message)
```
Lukasa, line 438: Minor nit: NIO style is four-space indent.

Sources/NIOSSH/SSHMessages.swift
```diff
@@ -419,6 +431,11 @@ extension ByteBuffer {
                 return .userAuthFailure(message)
             case SSHMessage.UserAuthSuccessMessage.id:
                 return .userAuthSuccess
+            case SSHMessage.UserAuthBannerMessage.id:
+              guard let message = self.readUserAuthBannerMessage()  else {
```
Lukasa, line 435:
```suggestion
              guard let message = self.readUserAuthBannerMessage() else {
```

Sources/NIOSSH/SSHMessages.swift
```diff
@@ -717,6 +734,18 @@ extension ByteBuffer {
...
+      self.rewindReaderOnNil { `self` in
+        guard let message = self.readSSHStringAsString(),
+              let languageTag = self.readSSHStringAsString()
+        else {
+          return nil
+        }
+
+        return SSHMessage.UserAuthBannerMessage(message: message, languageTag: languageTag)
+      }
+    }
```
Lukasa, line 747: Minor nit: NIO style is four-space indent.

Sources/NIOSSH/SSHMessages.swift
```diff
@@ -1296,6 +1328,13 @@ extension ByteBuffer {
         return writtenBytes
     }
 
+    mutating func writeUserAuthBannerMessage(_ message: SSHMessage.UserAuthBannerMessage) -> Int {
+      var writtenBytes = 0
+      writtenBytes += self.writeSSHString(message.message.utf8)
+      writtenBytes += self.writeSSHString(message.languageTag.utf8)
+      return writtenBytes
+    }
```
Lukasa, line 1336: Minor nit: NIO style is 4-space indent.

Sources/NIOSSH/SSHServerConfiguration.swift
```diff
@@ -23,9 +23,30 @@ public struct SSHServerConfiguration {
     /// The host keys for this server.
     public var hostKeys: [NIOSSHPrivateKey]
 
-    public init(hostKeys: [NIOSSHPrivateKey], userAuthDelegate: NIOSSHServerUserAuthenticationDelegate, globalRequestDelegate: GlobalRequestDelegate? = nil) {
+    /// The ssh banner to display to clients upon authentication
+    public var banner: UserAuthBanner?
+
+    public init(hostKeys: [NIOSSHPrivateKey], userAuthDelegate: NIOSSHServerUserAuthenticationDelegate, globalRequestDelegate: GlobalRequestDelegate? = nil, banner: UserAuthBanner? = nil) {
```
Lukasa, line 29: Sadly we can't add a new parameter to an existing function, even if it's optional: it's an API breaking change. We need to add a new `init` instead that accepts the new field.

Sources/NIOSSH/SSHServerConfiguration.swift
```diff
@@ -23,9 +23,30 @@ public struct SSHServerConfiguration {
...
         self.hostKeys = hostKeys
         self.userAuthDelegate = userAuthDelegate
         self.globalRequestDelegate = globalRequestDelegate ?? DefaultGlobalRequestDelegate()
+        self.banner = banner
+    }
+}
+
+// MARK: - UserAuthBanner
+public extension SSHServerConfiguration {
```
Lukasa, line 38: Minor nit: NIO style prefers to put the accessors on the things _within_ the extension, not the extension itself.

Sources/NIOSSH/SSHServerConfiguration.swift
```diff
@@ -23,9 +23,30 @@ public struct SSHServerConfiguration {
...
+     Client is obligated to display this banner to the end user, unless explicitely told
+     to ignore banners.
+     */
+    struct UserAuthBanner {
+        /// The message to display. Note that control characters contained in message
+        /// might be filtered in  accordance with the SSH specification.
+        let message: String
+
+        /// Language tag describing the language used for message. Must obey RFC 3066
+        let languageTag: String
```
Lukasa, line 50: I have the same type feedback here: can we make these `var` and provide a `public init`?

Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift
```diff
@@ -186,6 +186,10 @@ extension UserAuthenticationStateMachine {
             throw NIOSSHError.protocolViolation(protocolName: Self.protocolName, violation: "client sent user auth failure")
         }
     }
+
+    mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws {
+      // TODO: implement validation
```
Lukasa, line 191: Was this missed, or did you want to discuss it?

Lukasa (changes requested):
Excellent patch, thanks so much! The quality of this is really high, I've left a few notes within the diff but generally they're extremely small and nitpicky.

Regarding your discussion about when to send them, you're right that this is a little tricky. RFC 4252 definitely implies that the text is to be displayed _before_ authentication is attempted, which somewhat implies that we should take the opposite route and send it as soon as practical after the authentication protocol begins.

How important do you think the use-case of delaying this is? If it's vital, we can enable the functionality, but if you think it's mostly hypothetical then I think I'd lean towards the YAGNI principle and propose not enabling the functionality right now. We can always extend the config further to enable the functionality in a future patch if a use-case arises.

Lukasa:

Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift
DominikHorn: I indeed must have forgotten to implement this functionality. Thanks for the heads up

DominikHorn:
Thank you so much for the generous feedback, it is greatly appreciated :) 

I hope the patches I just submitted address your change requests adequately. Please let me know if there is anything else I may enhance in your opinion.

Regarding when to send the userauth banner: From my limited personal experience I would say that this is not terribly important. I have only encountered few servers that use this mechanism so far in the wild, and those that do seem to strictly use it to display hardware configuration information and administrator contact details. IMHO,  not leak userauth banner information to unauthenticated clients is therefore probably the right way to go in most real-world use case.

I believe that your suggestion to delay this additional functionality until it is actually requested is a great idea. Implementing it should be rather straight forward and only require another minor patch.

Lukasa (approved):
This is a really nice change, thanks so much for the work!

Lukasa:

Lukasa:
The soundness check has a few style changes that it would like you to make:

```diff
diff --git a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift
index de3dfc1..6bbb5d8 100644
--- a/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift	
+++ b/Sources/NIOSSH/Connection State Machine/Operations/AcceptsUserAuthMessages.swift	
@@ -67,7 +67,7 @@ extension AcceptsUserAuthMessages {
                 banner = config.banner
             }
 
-            return .possibleFutureMessage(future.map({ Self.transform($0, banner: banner)}))
+            return .possibleFutureMessage(future.map { Self.transform($0, banner: banner) })
         } else {
             return .noMessage
         }
@@ -92,8 +92,8 @@ extension AcceptsUserAuthMessages {
     }
 
     mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws -> SSHConnectionStateMachine.StateMachineInboundProcessResult {
-      try self.userAuthStateMachine.receiveUserAuthBanner(message)
-      return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
+        try self.userAuthStateMachine.receiveUserAuthBanner(message)
+        return .event(NIOUserAuthBannerEvent(message: message.message, languageTag: message.languageTag))
     }
 
     private static func transform(_ result: NIOSSHUserAuthenticationResponseMessage, banner: SSHServerConfiguration.UserAuthBanner? = nil) -> SSHMultiMessage {
diff --git a/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift b/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
index 32b56ce..f40925b 100644
--- a/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift	
+++ b/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift	
@@ -318,9 +318,9 @@ struct SSHConnectionStateMachine {
                 return result
 
             case .userAuthBanner(let message):
-              let result = try state.receiveUserAuthBanner(message)
-              self.state = .userAuthentication(state)
-              return result
+                let result = try state.receiveUserAuthBanner(message)
+                self.state = .userAuthentication(state)
+                return result
 
             case .disconnect:
                 self.state = .receivedDisconnect(state.role)
diff --git a/Sources/NIOSSH/SSHMessages.swift b/Sources/NIOSSH/SSHMessages.swift
index 4dcb037..45a0011 100644
--- a/Sources/NIOSSH/SSHMessages.swift
+++ b/Sources/NIOSSH/SSHMessages.swift
@@ -170,15 +170,15 @@ extension SSHMessage {
         static let id: UInt8 = 52
     }
 
-  struct UserAuthBannerMessage: Equatable {
-      // SSH_MSG_USERAUTH_BANNER
-      static let id: UInt8 = 53
+    struct UserAuthBannerMessage: Equatable {
+        // SSH_MSG_USERAUTH_BANNER
+        static let id: UInt8 = 53
 
-      /// message to display to user in client, encoded as ISO-10646 UTF-8 following RFC 3629
-      var message: String
+        /// message to display to user in client, encoded as ISO-10646 UTF-8 following RFC 3629
+        var message: String
 
-      /// tag identifying language of banner, following RFC 3066
-      var languageTag: String
+        /// tag identifying language of banner, following RFC 3066
+        var languageTag: String
     }
 
     struct UserAuthPKOKMessage: Equatable {
@@ -737,7 +737,7 @@ extension ByteBuffer {
     mutating func readUserAuthBannerMessage() -> SSHMessage.UserAuthBannerMessage? {
         self.rewindReaderOnNil { `self` in
             guard let message = self.readSSHStringAsString(),
-                  let languageTag = self.readSSHStringAsString()
+                let languageTag = self.readSSHStringAsString()
             else {
                 return nil
             }
diff --git a/Sources/NIOSSH/SSHServerConfiguration.swift b/Sources/NIOSSH/SSHServerConfiguration.swift
index 059d13e..fcd8ed0 100644
--- a/Sources/NIOSSH/SSHServerConfiguration.swift
+++ b/Sources/NIOSSH/SSHServerConfiguration.swift
@@ -39,6 +39,7 @@ public struct SSHServerConfiguration {
 }
 
 // MARK: - UserAuthBanner
+
 extension SSHServerConfiguration {
     /**
      Server sends `UserAuthBanner` to client some time during authentication.
@@ -50,7 +51,7 @@ extension SSHServerConfiguration {
          The message to be displayed by client to end user during authentication.
          Note that control characters contained in message might be filtered by
          client in accordance with RFC 4252.
-        */
+         */
         public var message: String
 
         /// Tag describing the language used for message. Must obey RFC 3066
diff --git a/Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift b/Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift
index bef3a4b..879548b 100644
--- a/Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift	
+++ b/Sources/NIOSSH/User Authentication/UserAuthenticationStateMachine.swift	
@@ -187,7 +187,7 @@ extension UserAuthenticationStateMachine {
         }
     }
 
-    mutating func receiveUserAuthBanner(_ message: SSHMessage.UserAuthBannerMessage) throws {
+    mutating func receiveUserAuthBanner(_: SSHMessage.UserAuthBannerMessage) throws {
         switch (self.delegate, self.state) {
         case (.client, .idle), (.client, .authenticationSucceeded):
             // Server sent a user auth success but we didn't ask them to!
diff --git a/Tests/NIOSSHTests/EndToEndTests.swift b/Tests/NIOSSHTests/EndToEndTests.swift
index ae6a198..e921eb8 100644
--- a/Tests/NIOSSHTests/EndToEndTests.swift
+++ b/Tests/NIOSSHTests/EndToEndTests.swift
@@ -130,7 +130,7 @@ struct TestHarness {
 
     var serverHostKeys: [NIOSSHPrivateKey] = [.init(ed25519Key: .init())]
 
-    var serverAuthBanner: SSHServerConfiguration.UserAuthBanner? = nil
+    var serverAuthBanner: SSHServerConfiguration.UserAuthBanner?
 }
 
 final class UserEventExpecter: ChannelInboundHandler {
diff --git a/Tests/NIOSSHTests/SSHMessagesTests.swift b/Tests/NIOSSHTests/SSHMessagesTests.swift
index 2c5111e..35f56c3 100644
--- a/Tests/NIOSSHTests/SSHMessagesTests.swift
+++ b/Tests/NIOSSHTests/SSHMessagesTests.swift
@@ -398,18 +398,18 @@ final class SSHMessagesTests: XCTestCase {
     }
 
     func testUserAuthBanner() throws {
-      var buffer = ByteBufferAllocator().buffer(capacity: 100)
-      let message = SSHMessage.userAuthBanner(.init(message: "Very important banner message containing crucial legal information.", languageTag: "en"))
-      
-      buffer.writeSSHMessage(message)
-      XCTAssertEqual(try buffer.readSSHMessage(), message)
-      
-      buffer.writeBytes([SSHMessage.UserAuthBannerMessage.id, 0, 0])
-      XCTAssertNil(try buffer.readSSHMessage())
-      
-      try self.assertCorrectlyManagesPartialRead(message)
-    }
-  
+        var buffer = ByteBufferAllocator().buffer(capacity: 100)
+        let message = SSHMessage.userAuthBanner(.init(message: "Very important banner message containing crucial legal information.", languageTag: "en"))
+
+        buffer.writeSSHMessage(message)
+        XCTAssertEqual(try buffer.readSSHMessage(), message)
+
+        buffer.writeBytes([SSHMessage.UserAuthBannerMessage.id, 0, 0])
+        XCTAssertNil(try buffer.readSSHMessage())
+
+        try self.assertCorrectlyManagesPartialRead(message)
+    }
+
     func testGlobalRequest() throws {
         let allocator = ByteBufferAllocator()
         var buffer = allocator.buffer(capacity: 100)
```

DominikHorn:
It was my pleasure, thanks for all the feedback :)

I hope I caught all formatting issues now. One interesting observation I made doing this: Within a clean repo at the current tip of main, the soundness.sh script reformats several lines throughout the project, e.g., moving accessor specifications from functions/variables to the enclosing (`extension`-) scope. Filtering just for lines touched by my commits, soundness.sh applies slightly different changes to the ones you cited in your previous message to some of the lines I added/changed with my commits. I therefore manually adjusted the formatting changes soundness.sh made locally to exactly match your requested changeset.

I suppose this could be a setup/versioning issue on my end. Just for future reference, e.g., if some else experiences this issue: I am using swift-format version `0.48.18` installed using arm64 homebrew on macOS Monterey.

Lukasa:
Yeah, swift-format is not stable across versions. I find this behaviour...frustrating.

Lukasa:

Lukasa:
Sadly one teeny tiny issue remains:

```diff
diff --git a/Sources/NIOSSH/SSHServerConfiguration.swift b/Sources/NIOSSH/SSHServerConfiguration.swift
index fadf945..fcd8ed0 100644
--- a/Sources/NIOSSH/SSHServerConfiguration.swift
+++ b/Sources/NIOSSH/SSHServerConfiguration.swift
@@ -39,6 +39,7 @@ public struct SSHServerConfiguration {
 }
 
 // MARK: - UserAuthBanner
+
 extension SSHServerConfiguration {
     /**
      Server sends `UserAuthBanner` to client some time during authentication.
```

Lukasa:

Lukasa:
I elected to just fix this up myself, as it was only a single line change and so well within the scope of something I can do in the web UI.

dnadoba:
AFAIK CI uses swift-format version `0.44.6`
https://github.com/apple/swift-nio-ssh/blob/09778e0388bda898c7592887f2ec84bb81ef21eb/docker/Dockerfile#L30
We should probably better document that in the README.md or CONTRIBUTING.md

PR #100, opened 2021-11-18
Generate up to 8 random padding bytes at a time instead of 4

gwynne:
Fixes a (presumably copy-pasta) typo in `ByteBuffer.writeSSHPaddingBytes(count:)` where only 4 random bytes would be requested from the `CSPRNG` at a time even when 8 or more were needed. The only impact of this issue is a very small performance penalty (so small that it's likely unmeasurable outside of microbenchmarks in practice, in fact), but might as well fix it anyhow.


Lukasa:

Lukasa (approved):

Lukasa (approved):

PR #101, opened 2021-11-18
Add `PeerMaximumMessageLength` option for SSH child channels.

### LINKED ISSUE #99: Child channels can not access their own peerMaxMessageSize
There does not appear to be any way for a child channel implementation to get access to the `peerMaxMessageSize` value derived from the `SSH_MESSAGE_CHANNEL_OPEN` message. This value is needed by some subsystems, such as SFTP, which are required to manually perform splitting of large data packets manually. In SFTP's case, offering an oversized value in a WriteFile request causes some servers to enter a wedged state waiting for data that will never arrive; without access to the channel's desired maximum message size, it must assume the worst (the specification-required minimum supported size of 32768 payload bytes), which is obviously less than ideal.

gwynne:
Fixes #99 (see issue for details).


Lukasa:

Lukasa (approved):
Nice change, LGTM!

Lukasa:
@gwynne Mind cleaning up the soundness check and updating to the tip of `main`?

PR #102, opened 2022-01-05
fixes an issue where decodeFirstBlock can be called multiple times

artemredkin:
Motivation:
When we call `decodeFirstBlock` we expect that first blockSize bytes,
that include packet length, will be overwritten in the parser buffer
(this is why we pass source byte buffer as inout). This changes state of
the parser, but if we cannot decode a packet we do not return it and in
the connection state machine we return without updating the machine
state. This can lead to calling decodeFirstBlock multiple times until we
have enough data to decode the whole packet in one go.

Modifications:
 - Update connection state machine state when active even if we do not
   have response message
 - Adds TinyEncriptionAlgorithm implementation to test codepaths that
   use decodeFirstBlock (unlike AESGCM)
 - Test that drip feeds data to parser

### DIFF
--- a/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
+++ b/Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
@@ -152,6 +152,7 @@ struct SSHConnectionStateMachine {
             }
         case .keyExchange(var state):
             guard let message = try state.parser.nextPacket() else {
+                self.state = .keyExchange(state)
                 return nil
             }
 
@@ -202,6 +203,7 @@ struct SSHConnectionStateMachine {
             }
         case .sentNewKeys(var state):
             guard let message = try state.parser.nextPacket() else {
+                self.state = .sentNewKeys(state)
                 return nil
             }
 
@@ -256,6 +258,7 @@ struct SSHConnectionStateMachine {
             // In this state we tolerate receiving service request messages. As we haven't sent newKeys, we cannot
             // send any user auth messages yet, so by definition we can't receive any other user auth message.
             guard let message = try state.parser.nextPacket() else {
+                self.state = .receivedNewKeys(state)
                 return nil
             }
 
@@ -287,6 +290,7 @@ struct SSHConnectionStateMachine {
         case .userAuthentication(var state):
             // In this state we tolerate receiving user auth messages.
             guard let message = try state.parser.nextPacket() else {
+                self.state = .userAuthentication(state)
                 return nil
             }
 
@@ -340,6 +344,7 @@ struct SSHConnectionStateMachine {
 
         case .active(var state):
             guard let message = try state.parser.nextPacket() else {
+                self.state = .active(state)
                 return nil
             }
 
@@ -405,6 +410,7 @@ struct SSHConnectionStateMachine {
             // We've received a key exchange packet. We only expect the first two messages (key exchange and key exchange init) before
             // we have sent a reply.
             guard let message = try state.parser.nextPacket() else {
+                self.state = .receivedKexInitWhenActive(state)
                 return nil
             }
 
@@ -449,6 +455,7 @@ struct SSHConnectionStateMachine {
         case .sentKexInitWhenActive(var state):
             // We've sent a key exchange packet. We expect channel messages _or_ a kexinit packet.
             guard let message = try state.parser.nextPacket() else {
+                self.state = .sentKexInitWhenActive(state)
                 return nil
             }
 
@@ -512,6 +519,7 @@ struct SSHConnectionStateMachine {
         case .rekeying(var state):
             // This is basically the regular key exchange state.
             guard let message = try state.parser.nextPacket() else {
+                self.state = .rekeying(state)
                 return nil
             }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
```diff
@@ -335,6 +335,7 @@ struct SSHConnectionStateMachine {
 
         case .active(var state):
             guard let message = try state.parser.nextPacket() else {
+                self.state = .active(state)
```
artemredkin, line 338: this is not the only place where we do not update the state, but it's the only one that breaks transport security schemes that use `decodeFirstBlock`. I also find it quite tricky to write a test for this. It should not affect initial key exchange since it's not encrypted, but it might affect re-keying. I will be trying to test that hypothesis, but wanted to start a conversation earlier in case you might come up with something as well...

Tests/NIOSSHTests/Utilities.swift
```diff
@@ -27,3 +29,224 @@ func assertNoThrowWithValue<T>(_ body: @autoclosure () throws -> T, defaultValue
         }
     }
 }
+
+// This algorithm is not secure, used only for testing purposes
+struct TinyEncryptionAlgorithm {
```
artemredkin, line 34: I'm not sure this is the best idea, but I don't think we can use swift-crypto here as it requires tag authentication, we need some encryption scheme were we can decode just the first block. Do you have any alternative ideas?

Tests/NIOSSHTests/SSHConnectionStateMachineTests.swift
```diff
@@ -512,6 +512,78 @@ final class SSHConnectionStateMachineTests: XCTestCase {
...
+        var server = SSHConnectionStateMachine(role: .server(.init(hostKeys: [NIOSSHPrivateKey(ed25519Key: .init())], userAuthDelegate: DenyThenAcceptDelegate(messagesToDeny: 0))), protectionSchemes: schemes)
+        try assertSuccessfulConnection(client: &client, server: &server, allocator: allocator, loop: loop)
+
+        let message = SSHMessage.channelData(.init(recipientChannel: 1, data: ByteBuffer(repeating: 17, count: 5)))
+
+        var tempBuffer = allocator.buffer(capacity: 1024)
+        XCTAssertNoThrow(try client.processOutboundMessage(message, buffer: &tempBuffer, allocator: allocator, loop: loop))
+        XCTAssert(tempBuffer.readableBytes > 0)
+
+        while var next = tempBuffer.readSlice(length: 1) {
```
artemredkin, line 572: the only alternative I was able to come up so far would be to expose state machine state and check if first block in the buffer is decoded, not sure if it's better

Tests/NIOSSHTests/Utilities.swift
Lukasa: Yeah, let's just do fixed-byte XOR. Encryption/decryption is XOR with a specific byte. We can then use SHA256 as a MAC. How does that sound?

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
Lukasa: Given that we know that `nextPacket` may mutate the state, we should probably unconditionally re-store the state in all locations, even when we can't easily write tests for this.

Lukasa: For re-keying I think you could try cribbing on the existing tests for re-keying.

Tests/NIOSSHTests/SSHConnectionStateMachineTests.swift
```diff
@@ -57,12 +64,80 @@ final class SSHConnectionStateMachineTests: XCTestCase {
             }
 
             if clientBuffer.readableBytes > 0 {
-                server.bufferInboundData(&clientBuffer)
+                if dripFeed {
```
Lukasa, line 67: Can we factor this block out to another function?

Tests/NIOSSHTests/SSHConnectionStateMachineTests.swift
```diff
@@ -57,12 +64,80 @@ final class SSHConnectionStateMachineTests: XCTestCase {
...
+                        if clientBuffer.readableBytes > 0 {
+                            switch try assertNoThrowWithValue(server.processInboundMessage(allocator: allocator, loop: loop)) {
+                            case .some(.emitMessage(let message)):
+                                serverMessage.append(message)
+                            case .none:
+                                ()
+                            case .some(.noMessage):
+                                ()
+                            case .some(.possibleFutureMessage(let futureMessage)):
+                                waitingForServerMessage = true
```
Lukasa, line 79: We need some caution here: can we `precondition` that `waitingForServerMessage` is `false` when we enter this block? The code doesn't handle it well.

Tests/NIOSSHTests/SSHConnectionStateMachineTests.swift
Lukasa: I don't think exposing the state machine state is the right approach here. This is probably fine.

Sources/NIOSSH/Connection State Machine/SSHConnectionStateMachine.swift
artemredkin: done

Tests/NIOSSHTests/SSHConnectionStateMachineTests.swift
artemredkin: done!

Lukasa (approved):

PR #106, opened 2022-03-30
Adopt `Sendable` on our public types

FranzBusch:
# Motivation
We need to adopt `Sendable` on certain types that are prone to be shared across threads.

# Modification
Adopt `Sendable` on some types.

# Result
Our types are `Sendable`

### DIFF
--- a/Sources/NIOSSH/Child Channels/SSHChannelData.swift
+++ b/Sources/NIOSSH/Child Channels/SSHChannelData.swift
@@ -12,7 +12,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if swift(>=5.6)
+@preconcurrency import NIOCore
+#else
 import NIOCore
+#endif // swift(>=5.6)
 
 /// `SSHChannelData` is the data type that is passed around in `SSHChildChannel` objects.
 ///
@@ -31,6 +35,8 @@ public struct SSHChannelData {
 }
 
 extension SSHChannelData: Equatable {}
+
+extension SSHChannelData: NIOSSHSendable {}
 
 extension SSHChannelData {
     /// The type of this channel data. Regular `.channel` data is the standard type of data on an `SSHChannel`,
@@ -57,6 +63,8 @@ extension SSHChannelData {
 }
 
 extension SSHChannelData.DataType: Hashable {}
+
+extension SSHChannelData.DataType: NIOSSHSendable {}
 
 extension SSHChannelData.DataType: CustomStringConvertible {
     public var description: String {
--- a/Sources/NIOSSH/Child Channels/SSHChildChannel.swift
+++ b/Sources/NIOSSH/Child Channels/SSHChildChannel.swift
@@ -26,7 +26,7 @@ import NIOConcurrencyHelpers
 /// for SSH purposes. In some cases it is possible to treat SSH channels as naive data pipes, but often the channels
 /// are intended for specific use-cases and need to be treated as special cases.
 final class SSHChildChannel {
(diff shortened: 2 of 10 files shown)


Sources/NIOSSH/Child Channels/SSHChannelData.swift
```diff
@@ -12,7 +12,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if swift(>=5.6)
+@preconcurrency import NIOCore
```
glbrntt, line 16: We should make a note somewhere to update this when NIO adds sendable checking.

Sources/NIOSSH/Child Channels/SSHChildChannel.swift
```diff
@@ -1184,3 +1188,16 @@ private extension IOData {
         }
     }
 }
+
+extension SSHChildChannel {
+    /// This is a ``Sendable`` box around a ``ChannelPipeline``. This is needed since the
+    /// ``SSHChildChannel`` needs to have an IUO to the pipeline because the pipeline needs a
```
glbrntt, line 1194: The `SSHChildChannel` is not marked as `Sendable`; should it be?

Lukasa: It's not clear to me why this is necessary: I think we should prefer `@unchecked Sendable` until such time as `ChannelPipeline` is made `Sendable`.

glbrntt: Even if `ChannelPipeline` is `Sendable`, `SSHChildChannel` would need to be `@unchecked Sendable` (because `var _pipeline: ChannelPipeline!` is mutable). 

I think the idea here was to explicitly mark this wrapper around `ChannelPipeline` as `@unchecked Sendable` so that everything else in `SSHChildChannel` has `Sendable` checks. Maybe that's overkill though 🤷‍♂️.

Lukasa: Hmm, I _suspect_ it's overkill, but I could be talked out of it.

FranzBusch: I see why it might be overkill but isn't it better to extract this single property out so that in the future if somebody adds something to the `SSHChildChannel` which would break sendability he will get a warning. Whereas if we make the whole channel `@unchecked Sendable` all future warnings will be silenced .

Sources/NIOSSH/Child Channels/SSHChannelData.swift
FranzBusch: Tracked in https://github.com/apple/swift-nio-ssh/issues/110

Sources/NIOSSH/Child Channels/SSHChildChannel.swift
Lukasa: I don't think @glbrntt's argument holds water, because the same argument applies to `PipelineSendableBox`: it has not magically made the mutability of the IUO go away. This appears to me to be "tricking" the compiler into thinking the property has suddenly become thread-safe, but the thread safety of this property has nothing to do with the box and everything to do with the fact that we don't publish a `self` reference anywhere except the `ChannelPipeline` before the `SSHChildChannel` is fully initialised: that is, we never write this field. I think the fact that this defeats `Sendable` checking is actually a compiler bug, and we should assume it will be fixed. This type does need to be `@unchecked Sendable`, because the compiler will struggle to prove that we're not doing anything dodgy with the IUO.

PR #109, opened 2022-04-03
Support identification string without carriage return

orobio:
Connecting to an OpenSSH 7.4 server fails due to that it only uses line
feeds and no carriage returns in the protocol version exchange.

This change makes sure that the identification string is properly
parsed when no carriage returns are present.

Since other clients have no problem connecting to OpenSSH 7.4, not
expecting a carriage return seems to be common. The following
from RFC 4253 also indicates that this is no problem:
"Implementers who wish to maintain compatibility with older,
undocumented versions of this protocol may want to process
the identification string without expecting the presence of
the carriage return character..."

### DIFF
--- a/Sources/NIOSSH/SSHPacketParser.swift
+++ b/Sources/NIOSSH/SSHPacketParser.swift
@@ -112,13 +112,22 @@ struct SSHPacketParser {
     }
 
     private mutating func readVersion() throws -> String? {
-        // Looking for a string ending with \r\n
-        let slice = self.buffer.readableBytesView
-        if let cr = slice.firstIndex(of: 13), cr.advanced(by: 1) < slice.endIndex, slice[cr.advanced(by: 1)] == 10 {
-            let version = String(decoding: slice[slice.startIndex ..< cr], as: UTF8.self)
-            // read \r\n
-            self.buffer.moveReaderIndex(forwardBy: slice.startIndex.distance(to: cr).advanced(by: 2))
-            return version
+        let carriageReturn = UInt8(ascii: "\r")
+        let lineFeed = UInt8(ascii: "\n")
+
+        // Search for version line, which starts with "SSH-". Lines without this prefix may come before the version line.
+        var slice = self.buffer.readableBytesView
+        while let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
+            if slice.starts(with: "SSH-".utf8) {
+                // Return all data upto the last LF we found, excluding the last [CR]LF.
+                slice = self.buffer.readableBytesView
+                let versionEndIndex = slice[lfIndex.advanced(by: -1)] == carriageReturn ? lfIndex.advanced(by: -1) : lfIndex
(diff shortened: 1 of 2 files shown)


Sources/NIOSSH/SSHPacketParser.swift
```diff
@@ -112,13 +112,22 @@ struct SSHPacketParser {
...
+        let carriageReturn = UInt8(ascii: "\r")
+        let lineFeed = UInt8(ascii: "\n")
+
+        // Search for version line, which starts with "SSH-". Lines without this prefix may come before the version line.
+        var slice = self.buffer.readableBytesView
+        while let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
+            if slice.starts(with: "SSH-".utf8) {
+                // Get all data upto the last line feed we found and filter out carriage returns.
+                slice = self.buffer.readableBytesView
+                let versionData = slice[slice.startIndex ..< lfIndex].filter { $0 != carriageReturn }
```
Lukasa, line 124: Why are we filtering out carriage returns?

Lukasa: I'm inclined to say that we should keep the carriage returns if they're present, and only drop the [CR]LF at the end.

orobio: It's updated, let me know what you think. Also, I don't have much experience with NIO/ByteBuffer[View], so please verify my logic there.

Sources/NIOSSH/SSHPacketParser.swift
```diff
@@ -112,13 +112,22 @@ struct SSHPacketParser {
...
+        while let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
+            if slice.starts(with: "SSH-".utf8) {
+                // Return all data upto the last LF we found, excluding the last [CR]LF.
+                slice = self.buffer.readableBytesView
+                let versionEndIndex = slice[lfIndex.advanced(by: -1)] == carriageReturn ? lfIndex.advanced(by: -1) : lfIndex
+                let version = String(decoding: slice[slice.startIndex ..< versionEndIndex], as: UTF8.self)
+                self.buffer.moveReaderIndex(forwardBy: slice.startIndex.distance(to: lfIndex).advanced(by: 1))
+                return version
+            } else {
+                slice = slice.dropFirst(slice.startIndex.distance(to: lfIndex.advanced(by: 1)))
```
Lukasa, line 129: We can probably clean this up a bit by doing:

```suggestion
                slice = slice[slice.index(after: lfIndex)...]
```

Lukasa (commented):
One really minor tweak here but otherwise this looks great!

Sources/NIOSSH/SSHPacketParser.swift
orobio: That definitely looks much better! It's updated.

Lukasa (approved):
Great change, thanks!

Lukasa:

PR #117, opened 2022-07-08
Use `swift-atomics` instead of `NIOAtomics`

dnadoba:
`NIOAtomics` was deprecated in https://github.com/apple/swift-nio/pull/2204 in favor of `swift-atomics` https://github.com/apple/swift-atomics
In addition, the `@preconcrrency` imports of `NIOCore` are not required and do not produce warnings/errors even with the currently latest released version of swift-nio (2.40.0), changes from `main` are not required.


glbrntt (approved):
LGTM, thanks David!

PR #118, opened 2022-08-01
Use Docc for documentation

Lukasa:
Motivation

Documentation is nice, and we can help support users by providing useful
clear docs.

Modifications

Add Docc to 5.6 and later builds
Make sure symbol references work
Add overview docs

Result

Nice rendering docs


FranzBusch (approved):
LGTM!

PR #119, opened 2022-08-11
fix typo in state machine type

JaapWijnen:
While working on the custom cryptography PR I ran across this small nit (easier to review this way)


JaapWijnen:
Hah, only now realise that Kex is shorthand for KeyExchange and isn't a typo 😆

PR #122, opened 2022-10-12
Replace `NIOSSHSendable` with `Sendable` and remove it

dnadoba:
We now require Swift 5.5.2 which includes `Sendable` so we no longer require the `NIOSSHSendable` workaround.
`NIOSSHSendable` is now no longer used and we can therefore remove it completely. This is would be a major change but we have not yet released a `1.x.x` version of `swift-nio-ssh`. I expect none of our users to relies on `NIOSSHSendable` conformance instead of `Sendable` so I hope this doesn't actually break any of our users.


Lukasa (approved):

Lukasa:
You might need to take a look at that 5.5 build failure.

dnadoba:
5.5 did catch a mistake! SSHChannelData is actually not `Sendable` because `IOData` is not `Sendable`. This produces only a warning in 5.5 because we have imported `NIOCore` with `@preconcurrency` in Swift 5.6+ and `IOData` was not marked explicitly non-`Sendable`. `IOData` and some other types will soon be marked explicitly non-`Sendable`: https://github.com/apple/swift-nio/pull/2290

Lukasa (approved):

PR #123, opened 2022-10-25
Refactoring of the encryptPacket method

artemredkin:
Motivation:
I think there is a bit of a mix of responsibilities between TransportProtection and PacketSerializer. PacketSerializer is responsible for writing out plaintext packet structure, but encrypted packet structure is written out by the TransportProtection implementations. This means that we will have duplicate implementations of basic packet writing logic, every implementation of TP will have to write length, padding length, payload and padding. Also, right now this API uses EncryptablePayload, which we cannot make fully public and accessible for writing tests as it will mean making all SSHMessages public as well.

Modifications:
 - Extract packet writing logic to a separate function that will be used by both plaintext and ciphertext modes
 - Removes usage of EncryptablePayload
 - Write plaintext packet structure (including payload) in the parser
 - TransportProtection implementations now only have to encrypt and tag

### DIFF
--- a/Sources/NIOSSH/SSHPacketSerializer.swift
+++ b/Sources/NIOSSH/SSHPacketSerializer.swift
@@ -47,40 +47,55 @@ struct SSHPacketSerializer {
                 preconditionFailure("only .version message is allowed at this point")
             }
         case .cleartext:
-            let index = buffer.writerIndex
-
-            /// Each packet is in the following format:
-            ///
-            ///   uint32        packet_length
-            ///   byte           padding_length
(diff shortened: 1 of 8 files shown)


Lukasa (approved):
Cool, this LGTM!

Lukasa:
This breaks our API, but we haven't gone 1.0 yet so this will just take us to 0.5.0.

artemredkin:
We made parts of `TransportProtection` APIs public with #97 and we haven't tagged anything yet, this only changes that part, yes

PR #124, opened 2022-11-02
Use #fileID/#filePath instead of #file

carolinacass:
Motivation:

#fileID introduced in Swift 5.3, so no longer need to use #file anywhere

Modifications:

Changed #file to #filePath or #fileID depending on the situation


glbrntt:

glbrntt (approved):
Thank you!

PR #127, opened 2022-12-08
Serialise SSH Public Key

carolinacass:
Motivation:
Needed a function function that takes a NIOSSHPublicKey and turns it into OpenSSH public key string


Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -56,6 +56,17 @@ public struct NIOSSHPublicKey: Hashable {
         self = key
     }
 
+
+    /// Takes a NIOSSHPublicKey and turns it into OpenSSH public key string
+    /// - Returns: OpenSSH public key string in the format of "algorithm-id base64-encoded-key"
+    public func keyIntoComponents() -> String {
```
Lukasa, line 62: So we should probably spell this as:

```swift
extension String {
    public init(openSSHPublicKey: NIOSSHPublicKey)
}
```

Lukasa (commented):
Generally looks really good! One minor nit on the API spelling, but otherwise this is fantastic.

Sources/NIOSSH/Keys And Signatures/NIOSSHPublicKey.swift
```diff
@@ -443,3 +443,13 @@ extension ByteBuffer {
         return returnValue
     }
 }
+extension String {
+    /// Takes a NIOSSHPublicKey and turns it into OpenSSH public key string in the format of "algorithm-id base64-encoded-key"
+    public init(openSSHPublicKey: NIOSSHPublicKey){
```
Lukasa, line 448:
```suggestion
    public init(openSSHPublicKey: NIOSSHPublicKey) {
```

Lukasa (approved):

argon:
🎉

PR #131, opened 2023-01-20
Remove some data-path CoW

Lukasa:
Motivation

CoW on the datapath is really gnarly, and we'd like to avoid it if at all possible.

Modifications

- Move some methods into mutating funcs on the state enum, which lets the compiler elide a retain/release pair that force a CoW
- Add proper reset code to a function that throws to ensure that the compiler can see that the state is always restored afterward.

Results

Drastically lowered allocation count


FranzBusch (approved):
I haven't the switch line by line because I assume this is just a copy and paste. Otherwise LGTM!

PR #132, opened 2023-01-23
Improve Sealed Box construction

Lukasa:
AES.GCM.SealedBox is expensive to construct from parts, in part because it's not inlinable. We can construct the combined representation directly and improve performance substantially by replacing 8 allocations with 3.


Sources/NIOSSH/TransportProtection/AESGCM.swift
```diff
@@ -342,3 +342,18 @@ extension Data {
         self = self[contentStartIndex ..< contentEndIndex]
     }
 }
+
+extension AES.GCM.SealedBox {
+    fileprivate init(nonce: SSHAESGCMNonce, ciphertext: ByteBufferView, tag: ByteBufferView) throws {
+        // As a workaround for a Swift Crypto inefficiency, we create the combined representation
+        // directly.
+        var combined: [UInt8] = []
+        combined.reserveCapacity(12 + ciphertext.count + tag.count)
```
dnadoba, line 351: can we add a comment that `12` is the size of `nonce` or can we make it a static property on `SSHAESGCMNonce`?

Lukasa: Done.

PR #139, opened 2023-04-12
Add docker-compose file for Swift 5.9

yim-lee:


yim-lee:
Also added CI job for `documentation-check`

Lukasa (commented):
I think you need to take the allocation counts from `main`, not from 5.8, as there was a regression on mainline we've yet to track down.

Lukasa (approved):

Issue #143, opened 2023-04-16
Crash in `SSHChildChannel`


gaetanzanella:
SwiftNIO SSH commit hash: c56ababb0e22e0cdf5ec8b31698e07f943ea92e7

## Context

While writing Xcode tests in one of my project which uses `swift-nio-ssh`, I encountered a crash.
I reproduced it using the `NIOSSHClient` target.

- [The code is available in my fork](https://github.com/gaetanzanella/swift-nio-ssh/pull/1/files)
- The crash happens when we force the closing of the channel while the command is running
- The crash only appears when using a command which produces a long output such as `yes \"long text\" | head -n 1000000\n`
- I only reproduced it when exchanging with the linked docker container. I'm using it in my automated tests.

## Steps to reproduce
1. Launch the ssh server: `docker-compose -f ./docker/ssh-docker-compose.yaml up -d`
2. Run the modified `NIOSSHClient` target. 
3. Observe the crash in Xcode
```
* thread #2, name = 'NIO-ELT-0-#0', stop reason = Fatal error: Sent channel window adjust on channel in invalid state
    frame #0: 0x0000000193fe4ae8 libswiftCore.dylib`_swift_runtime_on_report
    frame #1: 0x00000001940825d4 libswiftCore.dylib`_swift_stdlib_reportFatalErrorInFile + 208
    frame #2: 0x0000000193c614b4 libswiftCore.dylib`closure #1 (Swift.UnsafeBufferPointer<Swift.UInt8>) -> () in closure #1 (Swift.UnsafeBufferPointer<Swift.UInt8>) -> () in Swift._assertionFailure(_: Swift.StaticString, _: Swift.String, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never + 196
    frame #3: 0x0000000193c60210 libswiftCore.dylib`Swift._assertionFailure(_: Swift.StaticString, _: Swift.String, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never + 308
  * frame #4: 0x00000001002cae7c NIOSSHClient`ChildChannelStateMachine.sendChannelWindowAdjust(message=(recipientChannel = 0, bytesToAdd = 8811376), self=NIOSSH.ChildChannelStateMachine @ 0x0000000101e040a8) at ChildChannelStateMachine.swift:432:13
...
```
## Configuration
$ swift --version
[...]

PR #144, opened 2023-04-18
Avoid sending window update messages if locally closed (#143)

gaetanzanella:
Motivation:

Currently the library crashes if a window update is required and the connection is closed locally.

Modifications:

Check the connection state before sending a window update message.


Lukasa:

Lukasa (approved):
Awesome, thanks for this patch!

PR #146, opened 2023-05-30
Update update-alloc-limits-to-last-completed-ci-build

rnro:
Update to drop 55 support and add 59


glbrntt (approved):

Lukasa:
Looks like the nightly allocs have changed, can you run the new script?

PR #150, opened 2023-06-29
Correctly resize ByteBuffers

Lukasa:
Motivation:

In two different places in this module we "reserve" space by moving the writer index forward. This isn't the correct way to do that, and it's a bit of a needless performance
optimization. Instead, just write a zero in that location and come back to it.

Modifications:

- Replace unmotivated writer index move with a write.
- Add tests.

Result:

Fewer crashes

### DIFF
--- a/Sources/NIOSSH/SSHPacketSerializer.swift
+++ b/Sources/NIOSSH/SSHPacketSerializer.swift
@@ -73,7 +73,7 @@ extension ByteBuffer {
         ///   byte[m]   mac (Message Authentication Code - MAC); m = mac_length
 
         /// payload
-        self.moveWriterIndex(forwardBy: 5)
+        self.writeMultipleIntegers(UInt32(0), UInt8(0))
         let messageLength = self.writeSSHMessage(message)
 
         // Depending on on whether packet length is encrypted, padding should reflect that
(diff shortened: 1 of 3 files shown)


Sources/NIOSSH/SSHPacketSerializer.swift
```diff
@@ -73,7 +73,7 @@ extension ByteBuffer {
         ///   byte[m]   mac (Message Authentication Code - MAC); m = mac_length
 
         /// payload
-        self.moveWriterIndex(forwardBy: 5)
+        self.writeMultipleIntegers(UInt32(0), UInt8(0))
```
glbrntt, line 76: nit: there's also `writeRepeatedBytes` (or something similar to that) which might be a teeny tiny bit more obvious.

glbrntt (approved):

Sources/NIOSSH/SSHPacketSerializer.swift
Lukasa: Oh I actually thought this one made more sense given the packet format laid out above.

FranzBusch (approved):

Sources/NIOSSH/SSHPacketSerializer.swift
glbrntt: Makes sense, I was looking at it as the diff from “move the writer by 5”.

PR #151, opened 2023-06-30
Fully adopt `Sendable`

dnadoba:

PR #152, opened 2023-07-04
Update allocation counters

dnadoba:
Allocations have regressed without any code change. Likely cause is either swift (5.9 is still under development and I think 5.8 got a patch release) or one of our dependencies (swift-nio, swift-crypto, swift-atomics). We need to investigate why in the future.


Lukasa (approved):

dnadoba:
It appears that it is actually a regression of this PR: https://github.com/apple/swift-nio-ssh/pull/151

PR #153, opened 2023-07-21
Fixes client mode version parsing

artemredkin:
Motivation:
Server may send additional lines of data before version, but since we use version as part of key exchange, we need to filter out those lines, otherwise it will fail key exchange.

Modifications:
 - Strip out lines of data before version in client role
 - Update tests

### DIFF
--- a/Sources/NIOSSH/SSHPacketParser.swift
+++ b/Sources/NIOSSH/SSHPacketParser.swift
@@ -23,6 +23,7 @@ struct SSHPacketParser {
         case encryptedWaitingForBytes(UInt32, NIOSSHTransportProtection)
     }
 
+    private let isServer: Bool
     private var buffer: ByteBuffer
     private var state: State
     private(set) var sequenceNumber: UInt32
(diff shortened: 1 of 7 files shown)


Sources/NIOSSH/SSHPacketParser.swift
```diff
@@ -121,18 +123,31 @@ struct SSHPacketParser {
...
-        while let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
-            if slice.starts(with: "SSH-".utf8) {
-                // Return all data upto the last LF we found, excluding the last [CR]LF.
-                slice = self.buffer.readableBytesView
+        // Per RFC 4253 §4.2:
+        //  The server MAY send other lines of data before sending the version string.
+        // This means that server does not expect any lines before version so we will return all data before first line feed
+        if self.isServer {
+            // Looking for a string ending with \r\n
+            let slice = self.buffer.readableBytesView
```
Lukasa, line 131: Why are we dropping the requirement that the version line starts with `SSH-` on the server?

artemredkin: in server mode we require that client send no data before `SSH`, no need to look for it, we have a check later (and a test) that checks that version starts with `SSH-`

Lukasa (approved):

finagolfin:
@Lukasa, I see that the minimum NIO version needs to be bumped after this pull, as `takingOwnershipOfDescriptors()` added here was not added till NIO 2.56.0. This causes build failures for me locally because this package is checking out 2.45.0 instead.

PR #154, opened 2023-08-11
Fix some sendability warnings in tests

PeterAdams-A:
Motivation:

Swift concurrency warnings are here to help.
We can encode some of our assumptions to the type system.

Modifications:

Replace a number of captured variables with LoopBound instances.

Result:

Fewer sendability warnings


Lukasa (approved):

Lukasa:
The mighty power of `NIOLoopBoundBox` combining with `EmbeddedEventLoop`. I was confident this wouldn't work on first blush. Nice!

PR #164, opened 2023-12-18
Fix build on non-macOS Apple platforms

simonjbeaumont:
## Motivation

When building this project for Apple platforms other than macOS, the build failed for a couple of reasons. First, `NIOSSHServer` requires `Foundation.Process` which is only available on macOS. Second, the tests are making use of an CryptoKit API that's only present on iOS 13.2 (and aligned other versions), which is newer than the 13.0 stated in the Package.swift:

```
error: Cannot find type 'Process' in scope
...
error: 'isValidAuthenticationCode(_:authenticating:using:)' is only available in iOS 13.2 or newer
```

## Modifications

- Wrap all the code in `NIOSSHServer` with `#if canImport(Foundation.Process)` and, in `main.swift`, throw a `fatalError` with a helpful message.
- Add missing availability guards in test code. I've tried to keep them as scoped as possible.

## Result

Source and tests can now build on Apple platforms other than macOS.

### DIFF
--- a/Sources/NIOSSHServer/ExecHandler.swift
+++ b/Sources/NIOSSHServer/ExecHandler.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(Foundation.Process)
+
 import Dispatch
 import Foundation
 import NIOCore
@@ -156,3 +158,5 @@ final class ExampleExecHandler: ChannelDuplexHandler {
         }
     }
 }
+
+#endif // canImport(Foundation.Process)
--- a/Sources/NIOSSHServer/DataToBufferCodec.swift
+++ b/Sources/NIOSSHServer/DataToBufferCodec.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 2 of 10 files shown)


Sources/NIOSSHServer/ExecHandler.swift
```diff
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(Foundation.Process)
```
dnadoba, line 15: TIL that you can import check types and not only modules.

With that being said, why don't we use `@available` instead like we do in tests?

simonjbeaumont: Because that doesn't stop it trying to _compile_ the code. If `Foundation.Process` was _present_ on e.g. iOS but not _available_, then we could use `@available` here too, but it is simply not _present_ in the iOS SDK (and others), so we must use a compiler directive.

dnadoba: Makes sense! Thanks

dnadoba (approved):

Issue #170, opened 2024-09-29
Unable to cross compiling against MUSL.


zaneenders:
SwiftNIO SSH commit hash: e0ad5be6b474d24849b5b6ba8263bdfd0ede2eae

## Context:
Trying to compile `swift-nio-ssh` against MUSL getting the following errors. 
```
/Users/zane/Downloads/swift-nio-ssh/Sources/NIOSSH/Keys And Signatures/NIOSSHCertifiedPublicKey.swift:740:17: error: cannot find 'timespec' in scope
738 | extension DispatchWallTime {
739 |     init(secondsSinceEpoch: UInt64) {
740 |         let t = timespec(tv_sec: time_t(secondsSinceEpoch), tv_nsec: 0)
    |                 `- error: cannot find 'timespec' in scope
741 |         self = DispatchWallTime(timespec: t)
742 |     }

/Users/zane/Downloads/swift-nio-ssh/Sources/NIOSSH/Keys And Signatures/NIOSSHCertifiedPublicKey.swift:740:34: error: cannot find 'time_t' in scope
738 | extension DispatchWallTime {
739 |     init(secondsSinceEpoch: UInt64) {
740 |         let t = timespec(tv_sec: time_t(secondsSinceEpoch), tv_nsec: 0)
    |                                  `- error: cannot find 'time_t' in scope
741 |         self = DispatchWallTime(timespec: t)
742 |     }
```

## Steps to reproduce:
```
xcrun --toolchain swift swift build --swift-sdk aarch64-swift-linux-musl
# or 
xcrun --toolchain swift swift build --swift-sdk x86_64-swift-linux-musl
```

## Environment
```sh
xcrun --toolchain swift swift --version
Apple Swift version 6.0.1 (swift-6.0.1-RELEASE)
Target: arm64-apple-macosx15.0

xcrun --toolchain swift swift sdk list
swift-6.0.1-RELEASE_static-linux-0.0.1
```
Operating system: 15.0 (24A335)

PR #171, opened 2024-10-01
Add Musl support for static cross compilation.

zaneenders:
Motivation:

Provide a fix for Issue #170 as well as being able to use NIO SSH in projects that want to cross-compile to Linux.

Modifications:

Add import for Musl if available.

Result:

Cross-compiling to both x86_64 and aarch64 works.


Lukasa (approved):
Thanks! ✨

Lukasa:

PR #174, opened 2024-10-21
Migrate to GitHub Actions

rnro:
Migrate CI to use GitHub Actions with the exception of allocation count
performance tests which will remain on the old CI for now.

### Motivation:

To migrate to GitHub actions and centralised infrastructure.

### Modifications:

Changes of note:
* Disable warnings-as-errors until issues with recent NIO changes are
  resolved
* Adopt swift-format using rules from SwiftNIO
* Remove scripts which are no longer needed

### Result:

Feature parity with old CI, all-but-allocations migrated to GHA.


rnro:
@Lukasa When this PR is merged I we want to update the branch protection rules but in this case we want to retain the old CI PR validation checks because they (now only) run allocation count tests.

Lukasa (approved):

PR #183, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -81,3 +81,14 @@ let package = Package(
         ),
     ]
 )
+
+// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
+for target in package.targets {
+    if target.type != .plugin {
+        var settings = target.swiftSettings ?? []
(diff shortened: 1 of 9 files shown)


FranzBusch (approved):

finagolfin:
It appears this broke building with Swift 6.1 and [6.2 for linux](https://github.com/apple/swift-nio-ssh/actions/runs/12279462668/job/34263597139#step:4:680) and [Android](https://github.com/finagolfin/swift-android-sdk/actions/runs/12323129242/job/34398139212#step:29:1209). It looks like choosing to import `FoundationEssentials` _or_ `Foundation` in this pull was not the right approach, as `replacingOccurrences(of:with:options:range:)` is only available from the latter.

Lukasa:
Interesting, seems like it must have moved?

Lukasa:
See #185.

PR #190, opened 2025-03-03
Rename nightly_6_1 params to nightly_next

rnro:
Rename nightly_6_1 params to nightly_next; see https://github.com/apple/swift-nio/pull/3122


Lukasa (approved):
Manually validated the flags are being passed.

PR #200, opened 2025-03-31
Marks SSHChannelData as sendable

artemredkin:
Motivation:
Since IOData now sendable we can mark SSHChannel data sendable as well. This is required if we want to wrap SSH channel in a NIOAsyncChannel that operates with SSHChannelData

Modifications:
 - Remove unavailable from sendable conformance

Result:
 - SSHChannelData is now sendable


glbrntt (approved):
Nice one!

PR #210, opened 2025-06-18
Fix import guards in NIOSSHServer

rafaelcepeda:
The `NIOSSHServer` example guards all of its code with `#if canImport(Foundation.Process)` to avoid build failures on platforms without Process. 

However, Swift doesn't have submodules like this, Objective-C does. As Foundation transitions to Swift this check now always fails, even if Process is available. The result of this is that example code doesn't compile with newer Swift versions. Removing the can-imports results in a handful of Sendability warnings.

I've removed the guards and fixed all of the resulting Sendability warnings.


rafaelcepeda:
Closed this PR, as these changes have been incorporated in another PR.

PR #211, opened 2025-06-18
Fixed sendability warnings when -require-explict-sendable flag is enabled

rafaelcepeda:
Fixed sendability warnings when `-require-explict-sendable` flag is enabled.


Sources/NIOSSHServer/ExecHandler.swift
```diff
@@ -90,39 +91,42 @@ final class ExampleExecHandler: ChannelDuplexHandler {
...
-                let (ours, theirs) = GlueHandler.matchedPair()
-                try channel.pipeline.addHandler(ours).wait()
+        process.standardInput = inPipe
+        process.standardOutput = outPipe
+        process.standardError = errPipe
+        process.environment = self.environment
+        self.process = process
 
+        // Kick this off to a background queue
+        self.queue.sync {
```
glbrntt, line 119: This was previously `async` -- are we sure it's okay to block here?

Sources/NIOSSHServer/RemotePortForwarding.swift
```diff
@@ -26,16 +26,19 @@ import NIOSSH
 // Please note that, as with the rest of this example, there are important security features missing from
 // this demo.
 final class RemotePortForwarder {
-    private var serverChannel: Channel?
+    private let serverChannel: Channel?
```
glbrntt, line 29: Can we just delete this? It's always `nil`.

Sources/NIOSSHServer/RemotePortForwarding.swift
```diff
@@ -45,29 +48,33 @@ final class RemotePortForwarder {
...
                     )
-                )
-            ) { sshChildChannel, _ in
-                sshChildChannel.pipeline.addHandlers([DataToBufferCodec(), theirs]).flatMap {
-                    sshChildChannel.setOption(ChannelOptions.allowRemoteHalfClosure, value: true)
+                ) { sshChildChannel, _ in
+                    sshChildChannel.eventLoop.makeCompletedFuture {
+                        try sshChildChannel.pipeline.syncOperations.addHandlers([DataToBufferCodec(), theirs])
+                        _ = sshChildChannel.setOption(ChannelOptions.allowRemoteHalfClosure, value: true)
+                    }
```
glbrntt, line 72: This is returns a future so is async but we're ignoring it which means the this ssh child channel initializer is returning earlier than it should (i.e. before we know the option has been set).

You can do similar to what was done previously and `flatMap` the `setOption` call from the `makeCompletedFuture`:

```swift
sshChildChannel.eventLoop.makeCompletedFuture {
  // ...
}.flatMap { 
  sshChildChannel.setOption(ChannelOptions.allowRemoteHalfClosure, value: true)
}
```

Sources/NIOSSHServer/main.swift
```diff
@@ -62,18 +62,26 @@ defer {
...
     case .session:
-        return channel.pipeline.addHandler(ExampleExecHandler())
+        return channel.eventLoop.makeCompletedFuture {
+            try channel.pipeline.syncOperations.addHandler(ExampleExecHandler())
+        }
     case .directTCPIP(let target):
-        let (ours, theirs) = GlueHandler.matchedPair()
+        return channel.eventLoop.makeCompletedFuture {
+            let (ours, theirs) = GlueHandler.matchedPair()
+            _ = channel.pipeline.addHandler(DataToBufferCodec())
```
glbrntt, line 71: This should be a sync operation

Sources/NIOSSHServer/main.swift
```diff
@@ -62,18 +62,26 @@ defer {
...
+            try channel.pipeline.syncOperations.addHandler(ours)
+
+            let loopBoundHandler = NIOLoopBound(theirs, eventLoop: channel.eventLoop)
 
-        return channel.pipeline.addHandlers([DataToBufferCodec(), ours]).flatMap {
-            createOutboundConnection(
+            _ = createOutboundConnection(
```
glbrntt, line 76: We're ignoring the future here so the future returned from `sshChildChannelInitializer` can complete before all of the intended work is done. 

You'll need to chain this off of the `channel.eventLoop.makeCompletedFuture { ... }`

glbrntt (changes requested):

Sources/NIOSSHServer/ExecHandler.swift
rafaelcepeda: Good catch. This is not needed anymore, leftover from when I was testing different strategies. I've changed it back to `async`.

glbrntt (approved):
Looks great, thanks @rafaelcepeda!

PR #216, opened 2025-10-27
Bump minimum Swift version to 6.0

simonjbeaumont:
## Motivation

Now that Swift 6.2 has been released, we should drop support for 5.10, in line with our three version support window.

## Modifications

- Bump minimum tools version in package manifest
- Update README with minimum Swift version for each version range
- Remove compiler settings in package manifest that are subsumed by Swift 6 language mode
- Add `public typealias _NIOSSHSendableMetatype`, which resolves to `Swift.SendableMetatype` for Swift >= 6.2, and `Any` on < 6.2
- Have public protocols refine `SendableMetatype` on Swift >= 6.2, using `_NIOSSHSendableMetatype`

## Result

Package now requires Swift >= 6.0.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,8 @@
 name: Main
 
+permissions:
+    contents: read
+
 on:
     push:
         branches: [main]
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,5 +1,8 @@
 name: PR
 
+permissions:
+    contents: read
+
 on:
     pull_request:
         types: [opened, reopened, synchronize]
--- a/.github/workflows/pull_request_label.yml
+++ b/.github/workflows/pull_request_label.yml
@@ -1,5 +1,8 @@
 name: PR label
 
+permissions:
+  contents: read
+
 on:
   pull_request:
     types: [labeled, unlabeled, opened, reopened, synchronize]


Sources/NIOSSH/_NIOSSHSendableMetatype.swift
```diff
@@ -0,0 +1,19 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.2)
+@_marker public protocol _NIOSSHSendableMetatype: SendableMetatype {}
+#else
+@_marker public protocol _NIOSSHSendableMetatype {}
```
Lukasa: Don't introduce new protocols, use typealiases.

Package.swift
```diff
@@ -15,24 +15,6 @@
...
-    var initialSettings: [SwiftSetting] = []
-    initialSettings.append(contentsOf: [
-        .enableUpcomingFeature("StrictConcurrency"),
-        .enableUpcomingFeature("InferSendableFromCaptures"),
-    ])
-
-    if strictConcurrencyDevelopment {
-        // -warnings-as-errors here is a workaround so that IDE-based development can
-        // get tripped up on -require-explicit-sendable.
-        initialSettings.append(.unsafeFlags(["-Xfrontend", "-require-explicit-sendable", "-warnings-as-errors"]))
```
Lukasa: I think we still want the `-require-explicit-sendable` pieces, don't we?

Sources/NIOSSH/Constants.swift
```diff
@@ -15,7 +15,8 @@
 public enum Constants: Sendable {
     static let version = "SSH-2.0-SwiftNIOSSH_1.0"
 
-    public static let bundledTransportProtectionSchemes: [NIOSSHTransportProtection.Type] = [
-        AES256GCMOpenSSHTransportProtection.self, AES128GCMOpenSSHTransportProtection.self,
-    ]
+    public static let bundledTransportProtectionSchemes: [(NIOSSHTransportProtection & _NIOSSHSendableMetatype).Type] =
```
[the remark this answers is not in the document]
simonjbeaumont: Cool. I'll mark this one as resolved then.

PR #238, opened 2026-05-30
Fix readVersion() crash on bare LF as first byte (#237)

### LINKED ISSUE #237: Server crash: SSHPacketParser.readVersion() traps on bare LF as first byte
## Summary

`SSHPacketParser.readVersion()` in the server branch accesses `slice[lfIndex.advanced(by: -1)]` without checking that `lfIndex > slice.startIndex`. If the parser's `readableBytesView` begins with `\n` (e.g., the client sends `\n` as its first byte after the TCP handshake), the subscript indexes one byte before the buffer start and Swift's collection-bounds precondition fires, taking down the entire NIO event-loop thread with `EXC_BREAKPOINT`. Any unauthenticated TCP peer can crash the server with one byte.

Affects 0.12.0 and current `main` (verified by reading the source on both refs).

## Reproduction

Server-side: any NIOSSH server. Client-side:

```sh
printf '\n' | nc <server-host> <ssh-port>
```

The server process crashes with `EXC_BREAKPOINT (SIGTRAP)`.

## Crash stack

```
Thread 2 Crashed:: NIO-ELT-1-#1
0  libswiftCore.dylib  _assertionFailure(_:_:file:line:flags:) + 172
1  ByteBufferView.subscript.getter + 840         (ByteBuffer-views.swift:80)
2  SSHPacketParser.readVersion() + 520           (SSHPacketParser.swift:134)
3  SSHPacketParser.nextPacket() + 936            (SSHPacketParser.swift:71)
4  SSHConnectionStateMachine.State.processInboundMessage + 5584
5  SSHConnectionStateMachine.processInboundMessage + 96
6  NIOSSHHandler.channelRead(context:data:) + 708
…
```

## Offending code

[`Sources/NIOSSH/SSHPacketParser.swift`](https://github.com/apple/swift-nio-ssh/blob/0.12.0/Sources/NIOSSH/SSHPacketParser.swift#L122-L141), `readVersion()` server branch:

```swift
if self.isServer {
    let slice = self.buffer.readableBytesView
    if let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
        let versionEndIndex =
            slice[lfIndex.advanced(by: -1)] == carriageReturn ? lfIndex.advanced(by: -1) : lfIndex
        //   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ traps when lfIndex == slice.startIndex
        let version = String(decoding: slice[slice.startIndex..<versionEndIndex], as: UTF8.self)
        self.buffer.moveReaderIndex(forwardBy: slice.startIndex.distance(to: lfIndex).advanced(by: 1))
        return version
    }
}
```

The client branch below uses `slice.starts(with: "SSH-".utf8)` as a precondition and is not affected.

## Suggested fix

Guard the `-1` access:

```swift
if self.isServer {
    let slice = self.buffer.readableBytesView
    if let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
        let hasPriorByte = lfIndex > slice.startIndex
        let endsWithCR = hasPriorByte && slice[lfIndex.advanced(by: -1)] == carriageReturn
        let versionEndIndex = endsWithCR ? lfIndex.advanced(by: -1) : lfIndex
        let version = String(decoding: slice[slice.startIndex..<versionEndIndex], as: UTF8.self)
        self.buffer.moveReaderIndex(forwardBy: slice.startIndex.distance(to: lfIndex).advanced(by: 1))
        return version
    }
}
```

Stricter alternative: reject any banner that does not start with `SSH-` (server-side equivalent of the client branch's check), since a leading bare `\n` is not a legal SSH banner per RFC 4253 §4.2 anyway.

## Impact

Remote, pre-auth, unauthenticated DoS. One byte from any TCP peer terminates the entire server process — any other in-flight NIO connections handled by that event-loop group go down with it. We observed this in production after ~4 days uptime on a public SSH port (drive-by scanner traffic).

## Environment

- swift-nio-ssh: 0.12.0
- swift-nio: (latest compatible)
- Swift: 6.x
- macOS: 26.5

cyphers:
Closes #237.

`SSHPacketParser.readVersion()` in the server branch accessed
`slice[lfIndex.advanced(by: -1)]` without checking that
`lfIndex > slice.startIndex`. A client whose first byte after the TCP
handshake is a bare `\n` made the subscript index before the buffer
start, trapping the NIO event-loop thread with `EXC_BREAKPOINT` and
taking down the whole server process. Reproducible with
`printf '\n' | nc <host> <port>`.

## The fix

One-line guard on the `-1` access. A leading LF now yields an empty
version string, which is consistent with the existing server-branch
convention that everything before the first LF is the version line.

## Client branch

Unaffected — `slice.starts(with: \"SSH-\".utf8)` is checked first,
which guarantees `lfIndex >= 4 > slice.startIndex`.

## Test

Added `testReadVersionLineFeedFirstByteOnServer`. Verified that the
test reproduces the production crash signature
(`NIOCore/ByteBuffer-views.swift:80: Fatal error: index -1 out of range`)
without the fix, and passes with it. All 12 existing `SSHPacketParserTests`
still pass.

## Background

Reported in #237 after we hit this in production after ~4 days of
uptime on a public SSH port.

### DIFF
--- a/Sources/NIOSSH/SSHPacketParser.swift
+++ b/Sources/NIOSSH/SSHPacketParser.swift
@@ -130,8 +130,14 @@ struct SSHPacketParser {
             // Looking for a string ending with \r\n
             let slice = self.buffer.readableBytesView
             if let lfIndex = slice.firstIndex(of: lineFeed), lfIndex < slice.endIndex {
-                let versionEndIndex =
-                    slice[lfIndex.advanced(by: -1)] == carriageReturn ? lfIndex.advanced(by: -1) : lfIndex
+                // Guard the `-1` lookup: when the LF is the very first byte of
+                // the slice (e.g., a client sends a bare LF as its first byte
+                // after the TCP handshake), `lfIndex.advanced(by: -1)` would
+                // index before the buffer start and trap.
+                let endsWithCR =
+                    lfIndex > slice.startIndex
+                    && slice[lfIndex.advanced(by: -1)] == carriageReturn
+                let versionEndIndex = endsWithCR ? lfIndex.advanced(by: -1) : lfIndex
                 let version = String(decoding: slice[slice.startIndex..<versionEndIndex], as: UTF8.self)
                 self.buffer.moveReaderIndex(forwardBy: slice.startIndex.distance(to: lfIndex).advanced(by: 1))
                 return version
--- a/Tests/NIOSSHTests/SSHPacketParserTests.swift
+++ b/Tests/NIOSSHTests/SSHPacketParserTests.swift
@@ -134,6 +134,27 @@ final class SSHPacketParserTests: XCTestCase {
         }
     }
 
+    func testReadVersionLineFeedFirstByteOnServer() throws {
+        // Regression test for the crash described in #237. SSHPacketParser
+        // previously accessed `slice[lfIndex.advanced(by: -1)]` without
+        // checking that `lfIndex > slice.startIndex`, so a client sending a
+        // bare LF as its first byte after the TCP handshake trapped the
+        // process. Per the existing server-branch semantics (everything
+        // before the first LF is the version line), a leading LF should
+        // simply yield an empty version string and not crash.
+        var parser = SSHPacketParser(isServer: true, allocator: ByteBufferAllocator())
+
+        var part1 = ByteBuffer.of(string: "\n")
+        parser.append(bytes: &part1)
+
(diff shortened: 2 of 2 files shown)


josephnoir:
LGTM. Thank you for the fix!

josephnoir (approved):

PR #240, opened 2026-06-12
add agent forwarding support

zg:


Lukasa (changes requested):
Thanks for this, do you mind writing some tests as well?

zg:
> Thanks for this, do you mind writing some tests as well?

Done - mind taking another look?

PR #245, opened 2026-07-20
Configurable max packet size

josephnoir:
### Motivation

The advertised max packet size was not configurable and used a
relatively high hard-code default.

### Modifications

* Configurable max packet size and lower default.
* Base window size on the configured max packet size to allow space for 64 packets.

### Results

More configurability.

### DIFF
--- a/Sources/NIOSSH/SSHClientConfiguration.swift
+++ b/Sources/NIOSSH/SSHClientConfiguration.swift
@@ -36,6 +36,7 @@ public struct SSHClientConfiguration {
     ///
     /// - Precondition: Must be at least 32768 bytes, the uncompressed payload size
     ///   RFC 4253 §6.1 requires every implementation to support.
+    /// - Precondition: The packet size must leave 1024 bytes for headers and framing.
     public var maximumPacketSize: Int = Constants.defaultMaximumChannelPacketSize {
         didSet {
             precondition(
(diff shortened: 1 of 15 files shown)


Lukasa (approved):

Sources/NIOSSH/SSHClientConfiguration.swift
```diff
@@ -26,6 +26,30 @@ public struct SSHClientConfiguration {
     /// Supported data encryption algorithms
     public var transportProtectionSchemes: [NIOSSHTransportProtection.Type]
 
+    /// The maximum size, in bytes, of a channel data payload this peer is willing to receive (the
+    /// "maximum packet size" of an SSH channel, RFC 4254 §5.1). It is advertised to the remote peer
+    /// when opening channels and bounds inbound encrypted packets. Defaults to `1 << 17`.
```
glbrntt: nit: might be worth stating what `1 << 17` is 128 KiB.

Sources/NIOSSH/SSHServerConfiguration.swift
```diff
@@ -29,6 +29,30 @@ public struct SSHServerConfiguration {
     /// Supported data encryption algorithms
     public var transportProtectionSchemes: [NIOSSHTransportProtection.Type]
 
+    /// The maximum size, in bytes, of a channel data payload this peer is willing to receive (the
+    /// "maximum packet size" of an SSH channel, RFC 4254 §5.1). It is advertised to the remote peer
+    /// when opening channels and bounds inbound encrypted packets. Defaults to `1 << 17`.
```
glbrntt: same comment re: 128 KiB

glbrntt (approved):

Sources/NIOSSH/SSHClientConfiguration.swift
josephnoir: Good idea, added.

